PR #47, opened 2017-03-15
PP-1718 Pack Provider tests (User Api)

rauligs:
## WHAT
- State data for create user when username already exists (unhappy scenario).
- Create a base user as username 'existing-user' for idempotent API operations.
- Add state for forgotten password scenarios as Consumer expects (fixed codes).
- Add state for login attempt scenario (max reached).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,11 @@
             <version>2.0.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>au.com.dius</groupId>
+            <artifactId>pact-jvm-provider-junit_2.11</artifactId>
+            <version>3.3.1</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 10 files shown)


mrlumbu:
Maybe change titile of PR to Pact Provider tests?

mrlumbu:
Just a comment: need to decide whether we want to continue to commit pact files (until we have a more sustainable solution)

rauligs:
You are right. For now we would have two different scenarios,
- **Changes in the Provider _(Adminusers)._**. If something changes in the API these tests will fail so we need at least to make sure we are in sync with the Consumers expectations unless they changed (so they generate new files that need to be copied across).
 - **Changes in the Consumer _(Selfservice)._** At the moment probably worth to keep them and make sure we commit new files in _adminusers_ when _selfservice_ expectations change (and generates new _pact_ files), copy them across.
- All of this will go away as soon we have Pact broker and the CI ready for contract testing, but for now a good exercise to learn more about _Pact._

mrlumbu:
Yep, sounds sensible 👍

PR #50, opened 2017-03-17
PP-1716 adjust 2fa login counter check

kakumara:
The way selfservice checks for user is disabled or not at first factor and second factor are slightly different. 
On first factor (username/password) it checks disability before authenticating username/password. On Second factor we check after 2fa verification. 
So, in order to have the same locking effect to user we need to adjust the 2fa locking set only after max_attempts exceeded.


SandorArpa:
LGTM

PR #51, opened 2017-03-20
PP-1716 Remove login attempts api's no longer in use

kakumara:
2FA now happens in adminusers. We no longer need to expose these API's to outside.


kakumara:
👍  updated readme and removed few other unused references.

PR #54, opened 2017-03-22
PP-1819 update user service role

kakumara:
Introduced `PUT  /v1/api/users/{username}/services/{service-id}`

Instead of adding business logic on to the existing UserService (which is already grown bit too much), have introduced a guice assisted factory pattern to obtain business service components that can be used by UserResource. The upside of this guice factory is that we can have @Transactional around the component method as usual.
And we can unit test each individual business service object separately without messing around other business services (unlike UserServicesTest.java)

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceRoleUpdater.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceRoleUpdater.java
@@ -1,0 +1,74 @@
+package uk.gov.pay.adminusers.service;
+
+import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
+import uk.gov.pay.adminusers.model.User;
+import uk.gov.pay.adminusers.persistence.dao.RoleDao;
+import uk.gov.pay.adminusers.persistence.dao.ServiceDao;
+import uk.gov.pay.adminusers.persistence.dao.UserDao;
+import uk.gov.pay.adminusers.persistence.entity.RoleEntity;
+import uk.gov.pay.adminusers.persistence.entity.ServiceRoleEntity;
+import uk.gov.pay.adminusers.persistence.entity.UserEntity;
+
+import java.util.Optional;
+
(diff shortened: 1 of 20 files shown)


src/main/java/uk/gov/pay/adminusers/service/ServiceRoleUpdater.java
```diff
@@ -0,0 +1,74 @@
...
+        if(!servicesRoleMaybe.isPresent()) {
+            throw conflictingServiceForUser(userEntity.getId(), serviceId);
+        }
+
+        ServiceRoleEntity serviceRoleEntity = servicesRoleMaybe.get();
+
+        if (!roleEntity.getId().equals(ROLE_ADMIN_ID)) {
+            if (serviceDao.countOfRolesForService(serviceId, ROLE_ADMIN_ID) <= adminsPerServiceLimit) {
+                throw adminRoleLimitException(adminsPerServiceLimit);
+            }
```
mrlumbu: is this right? a hardcoded limit of 1 admin per service? I think may of our users already violate that

mrlumbu: sorry read it wrong - its a minimum limit

kakumara: That is the requirement (at least 1 admin required). Yes/Maybe about the existing services, but at the moment this check is enforced only at change of user role. So they can still exist from provisioning.

mrlumbu:
I think this looks fine. Just as an FYI for anyone else: Discussed with Kasun the use of guice assisted inject versus partitioning up resource, and figure this is fine for now but we should keep an eye on resource becoming too big (or getting too many cross dependencies), and maybe review then, but at the moment this is a good approach to cleanly modularise different pieces of business logic, and we should add more components to the UserServicesFactory in future stories

PR #55, opened 2017-03-22
PP-1717 use external id not username in api urls

DanailMinchev:
## WHAT

We are identifying user resource based on `username`. Currently this is also the email address of the user. As it is in the resource path this is being logged everywhere. We need to introduce a separate identifier.

## HOW

- Added new random UUID generator
- Added `external_id` database column and modified related code
- API related changes
- Added documentation

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/ForgottenPassword.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/ForgottenPassword.java
@@ -18,32 +18,26 @@ public class ForgottenPassword {
     private Integer id;
     private String code;
     private ZonedDateTime date;
-    private String userExternalId;
     private String username;
     private List<Link> links;
 
-    public static ForgottenPassword forgottenPassword(String code, String userExternalId, String username) {
-        return forgottenPassword(randomInt(), code, userExternalId, username, ZonedDateTime.now(ZoneId.of("UTC")));
+    public static ForgottenPassword forgottenPassword(String code, String username) {
+        return forgottenPassword(randomInt(), code, username, ZonedDateTime.now(ZoneId.of("UTC")));
     }
 
-    public static ForgottenPassword forgottenPassword(Integer id, String code, String userExternalId, String username, ZonedDateTime date) {
-        return new ForgottenPassword(id, code, date, userExternalId, username);
+    public static ForgottenPassword forgottenPassword(Integer id, String code, String username, ZonedDateTime date) {
+        return new ForgottenPassword(id, code, date, username);
     }
 
-    private ForgottenPassword(Integer id, String code, ZonedDateTime date, String userExternalId, String username) {
+    private ForgottenPassword(Integer id, String code, ZonedDateTime date, String username) {
         this.id = id;
         this.code = code;
         this.date = date;
-        this.userExternalId = userExternalId;
         this.username = username;
     }
 
     public Integer getId() {
         return id;
-    }
-
-    public String getUserExternalId() {
-        return userExternalId;
     }
 
     public String getCode() {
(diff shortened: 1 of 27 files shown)


src/main/java/uk/gov/pay/adminusers/model/ForgottenPassword.java
```diff
@@ -18,28 +18,34 @@
     private Integer id;
     private String code;
     private ZonedDateTime date;
+    private String userExternalId;
```
kakumara: I doubt we ever need this attribute in `ForgottenPassword`. We should be able to do forgotten password without this thing

src/main/java/uk/gov/pay/adminusers/resources/ForgottenPasswordResource.java
```diff
@@ -39,11 +39,14 @@ public ForgottenPasswordResource(ForgottenPasswordServices forgottenPasswordServ
     public Response createForgottenPassword(JsonNode payload) {
         logger.info("ForgottenPassword CREATE request - [ {} ]", payload);
         Optional<Errors> errorsOptional = validator.validateCreateRequest(payload);
+        final String usernameOrExternalId = payload.get("username") != null ?
+                payload.get("username").asText() :
+                payload.get("external_id").asText();
```
kakumara: Once again, I dont think we will ever need this

src/main/java/uk/gov/pay/adminusers/service/ForgottenPasswordServices.java
```diff
@@ -29,15 +30,19 @@ public ForgottenPasswordServices(UserDao userDao, ForgottenPasswordDao forgotten
     }
 
     @Transactional
-    public Optional<ForgottenPassword> create(String username) {
-        return userDao.findByUsername(username)
+    public Optional<ForgottenPassword> create(String usernameOrExternalId) {
```
kakumara: same as above

src/test/java/uk/gov/pay/adminusers/service/UserServicesTest.java
```diff
@@ -259,12 +284,12 @@ public void shouldReturnUser_whenIncrementingSessionVersion_ifUserFound() throws
         UserEntity userEntity = aUserEntityWithTrimmings(user);
 
         Optional<UserEntity> userEntityOptional = Optional.of(userEntity);
-        when(userDao.findByUsername("random-name")).thenReturn(userEntityOptional);
+        when(userDao.findByUsername(USER_USERNAME)).thenReturn(userEntityOptional);
 
-        Optional<User> userOptional = userServices.patchUser("random-name", PatchRequest.from(node));
+        Optional<User> userOptional = userServices.patchUser(USER_USERNAME, PatchRequest.from(node));
```
kakumara: We should possibly start adopting EXTERNAL_IDs for these methods (instead of USER_USERNAME). But can be done in a next PR if needed.

DanailMinchev: As discussed I will change that in the next PR

heathd:
@DanailMinchev hey Dan, just a small improvement suggestion. You have really good commit messages in here but they're in the commit body rather than the title (first line of the commit message), so they're not so easily visible.

It's preferable to make the first line of the commit message a succinct summary of the change, and use the commit body to elaborate and explain any further details, rationale etc if needed.

For example for 6d81b32 I would suggest a commit message of:

```
PP-1717 Added new random UUID generator
```

For other commits you can probably add a first line which summarises the commit.

For c8ed29e you have two bullet points in the message body which suggests that it might be better to split into two commits for each of the two changes.

PR #56, opened 2017-03-24
PP-1605 Order users by username rather than role.

rauligs:
## WHAT
- Order users by username rather than role.


rauligs:
rebuild

PR #60, opened 2017-04-04
PP-1896 Username and email case insensitive on findBy DAO methods

rauligs:
- We have an issue in forgotten passwords not resetting passwords when email does not match db case so making the find case insensitive for both username (is this what is being used) and email.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
@@ -31,21 +31,21 @@ public class UserDao extends JpaDao<UserEntity> {
 
     public Optional<UserEntity> findByUsername(String username) {
         String query = "SELECT u FROM UserEntity u " +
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
```diff
@@ -31,21 +31,21 @@ public UserDao(Provider<EntityManager> entityManager) {
 
     public Optional<UserEntity> findByUsername(String username) {
         String query = "SELECT u FROM UserEntity u " +
-                "WHERE u.username = :username";
+                "WHERE LOWER(u.username) = :username";
```
heathd: might be better to use PG's `lower()` function on both strings to avoid the possibility that e.g. unicode chars get treated differently by the different implementations

rauligs: @heathd Very good point, I missed it, will create another PR to do this

PR #66, opened 2017-04-11
PP-1717 Genarate externalId only when needed

DanailMinchev:
- Removed random exteranlId generation from JsonNode: we must ignore JsonNode externalIds (coming from clients) when creating User object
- Generate externalIds using Java


DanailMinchev:
rebuild

PR #69, opened 2017-04-26
PP-1603 Invite Users API

rauligs:
### Create invites
- Add migration script for invites
- Add sender of the invite as foreign key _users(id)_
- Validate invites includes payload validations and sender validations (such as the sender must be and Admin of the given service).
- Async notification with invitation link pointing to selfservice. No guarantee then that the email is sent but still created the invite.
- It can be created multiple invites to the same email at the moment. This might change (performing an override).

### Get invites (validation)
- Performs validation of the given invitation code (in url set by create invites)
- Check for expiry (next day midnight of the invite date). See tests for documentation.
- It returns email and telephone_number if exists of the user being invited (this is because this URL can be accessed multiple times, and the user could set already its telephone number - in next resource - adding telephone_number and password when generating OTP).


src/main/java/uk/gov/pay/adminusers/persistence/dao/RoleDao.java
```diff
@@ -29,9 +29,11 @@ public RoleDao(Provider<EntityManager> entityManager){
         String query = "SELECT r FROM RoleEntity r " +
                 "WHERE r.name = :roleName";
 
-        return entityManager.get()
+        Optional<RoleEntity> roleName1 = entityManager.get()
```
SandorArpa: Forgotten to revert debug changes?

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -1,30 +1,48 @@
...
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.Status.*;
+import static javax.ws.rs.core.UriBuilder.fromUri;
 
 @Path("/v1/api/services")
 public class ServiceResource {
 
-    private static final Logger logger = PayLoggerFactory.getLogger(ServiceResource.class);
+    private static final Logger LOGGER = PayLoggerFactory.getLogger(ServiceResource.class);
+    private final String selfServiceUrl;
```
SandorArpa: What is it doing here?

src/main/java/uk/gov/pay/adminusers/service/ServiceRoleUpdater.java
```diff
@@ -64,8 +64,8 @@ public ServiceRoleUpdater(UserDao userDao, ServiceDao serviceDao, RoleDao roleDa
 
         ServiceRoleEntity serviceRoleEntity = servicesRoleMaybe.get();
 
-        if (!roleEntity.getId().equals(ROLE_ADMIN_ID)) {
-            if (serviceDao.countOfRolesForService(serviceId, ROLE_ADMIN_ID) <= adminsPerServiceLimit) {
+        if (!roleEntity.isAdmin()) {
```
SandorArpa: Bit confusing. If NOT admin then check for admin limits?

rauligs: As discussed, yes is a bit confusing thinking about this situation from a _selfservice_ as client POV but makes sense if the role to changes is something else but Admin since we can't left a service empty of Admin users (at least one must exist)

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
rauligs: Will remove it, refactoring made not in use.

src/main/java/uk/gov/pay/adminusers/persistence/dao/RoleDao.java
rauligs: 👍 will change this

src/main/java/uk/gov/pay/adminusers/service/ServiceRoleUpdater.java
SandorArpa: 👍

PR #71, opened 2017-04-27
PP-1348 make contract tests more flexible

mrlumbu:
In order to be able to run contract tests against a broker we need to be able to configure how contract tests get hold of contracts at run time. This work enables that by introducing a `ConfigurablePactLoader` which can load pacts either from local files or from pact broker. At the moment by default it loads from local, which means all tests pass without modification.

While we have tested whether this loader does work with a remote broker, I would rather ignore whether that bit works for now and just get this in. When we do need it to work with a remote broker we can make any changes necessary.


pom.xml
```diff
@@ -208,7 +208,7 @@
         <dependency>
             <groupId>au.com.dius</groupId>
             <artifactId>pact-jvm-provider-junit_2.11</artifactId>
-            <version>3.3.1</version>
+            <version>3.3.7</version>
```
DanailMinchev: I can see there is a new `3.3.8` release (https://github.com/DiUS/pact-jvm/releases), should we update it?

mrlumbu: There does not seem to be a compelling reason to go to 3.3.8 right now, so lets do this separately.

PR #74, opened 2017-05-02
PP-1977 Forgotten Password notifications

rauligs:
## WHAT
- Added notifications as part of forgotten password journey. Since we don't need any information to send back to selfservice (was building the link for the notification in that side), removed response payload, returning a 200 OK response when creating a Forgotten Password successfully (consistent with second factor)

- These changes assume selfservice is using v2 of the resource. This PR will make possible then to use this new version (next PR). So this PR depends on https://github.com/alphagov/pay-selfservice/pull/332 to be merged and deployed.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
+++ b/src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
@@ -16,8 +16,6 @@ import uk.gov.pay.adminusers.validations.RequestValidations;
 
 import java.util.Properties;
 
-import static uk.gov.pay.adminusers.app.util.TrustStoreLoader.getSSLContext;
-
 public class AdminUsersModule extends AbstractModule {
 
     final AdminUsersConfig configuration;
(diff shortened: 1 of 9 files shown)


kakumara:
V2?  I kind of recall we had a long debate over V2 and maybe decided not to version internal APIs??cc: @mrlumbu

kakumara:
Other than that 👍

kakumara:
pay

src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
```diff
@@ -77,6 +77,7 @@ private NotificationService provideUserNotificationService() {
                 new NotifyClientProvider(configuration.getNotifyConfiguration(), getSSLContext()),
                 configuration.getNotifyConfiguration().getSecondFactorSmsTemplateId(),
                 configuration.getNotifyConfiguration().getInviteEmailTemplateId(),
+                configuration.getNotifyConfiguration().getForgottenPasswordEmailTemplateId(),
```
DanailMinchev: Addressed in https://github.com/alphagov/pay-adminusers/pull/74/commits/ffddd2eb0c6381b040f23dd4d6a422a1be849a3a

DanailMinchev:
About the `V2` versioning comment:
as discussed this is only temporary, we will use `v1` after clean up

PR #80, opened 2017-05-15
PP-1603 Invite Resource OTP

DanailMinchev:
## WHAT

- Add invites expiry date. So move logic into the entity.
- Refactored invites otp resources paths.
- Add validation for otp requests.
- Add resource to create user on otp validation.
- Add resource to resend otp
- Return newly created User object on successful invite validation

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/User.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/User.java
@@ -18,10 +18,8 @@ public class User {
     public static final String FIELD_USERNAME = "username";
     public static final String FIELD_PASSWORD = "password";
     public static final String FIELD_EMAIL = "email";
-    public static final String FIELD_SERVICE_IDS = "service_ids";
     public static final String FIELD_GATEWAY_ACCOUNT_IDS = "gateway_account_ids";
     public static final String FIELD_TELEPHONE_NUMBER = "telephone_number";
-    public static final String FIELD_OTP_KEY = "otp_key";
     public static final String FIELD_ROLE_NAME = "role_name";
 
     private Integer id;
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.adminusers.persistence.entity;
 
+import uk.gov.pay.adminusers.app.util.RandomIdGenerator;
 import uk.gov.pay.adminusers.model.Invite;
 
 import javax.persistence.*;
@@ -8,7 +9,6 @@ import java.time.ZonedDateTime;
 
 import static java.time.ZonedDateTime.now;
 import static java.time.temporal.ChronoUnit.DAYS;
-import static uk.gov.pay.adminusers.app.util.RandomIdGenerator.newId;
 import static uk.gov.pay.adminusers.persistence.entity.UTCDateTimeConverter.UTC;
 
 @Entity
@@ -20,6 +20,10 @@ public class InviteEntity extends AbstractEntity {
     @Column(name = "date")
     @Convert(converter = UTCDateTimeConverter.class)
     private ZonedDateTime date;
+
+    @Column(name = "expiry_date")
+    @Convert(converter = UTCDateTimeConverter.class)
+    private ZonedDateTime expiryDate;
 
     @ManyToOne
     @JoinColumn(name = "role_id")
@@ -52,14 +56,33 @@ public class InviteEntity extends AbstractEntity {
         //for jpa
     }
 
-    public InviteEntity(String email, String code, UserEntity sender, ServiceEntity service, RoleEntity role) {
+    public InviteEntity(String email, String code, String otpKey, UserEntity sender, ServiceEntity service, RoleEntity role) {
         this.service = service;
         this.date = now(ZoneId.of("UTC"));
+        initializeExpiry();
         this.code = code;
-        this.otpKey = newId();
+        this.otpKey = otpKey;
         this.sender = sender;
         this.email = email.toLowerCase();
         this.role = role;
+    }
+
+    /**
+     * Being:
+     * <p>
+     * 'X' the moment the invite is created and
+     * '|' = 00:00 of the following day
+     * '^' the moment it expires
+     * 'N' = Now
+     * <p>
+     * <-------Day 0---------><-------Day 1---------><-------Day 2--------->
+     * |----------------------|----------------------|----------------------|
+     * X                               N             ^
+     * <p>
+     * Invite created Day 1 -> 00:00:00:000 will expired at Day 3 -> 00:00:00:000
+     */
+    private void initializeExpiry() {
+        this.expiryDate = this.date.truncatedTo(DAYS).plus(EXPIRY_DAYS, DAYS);
     }
 
     public ZonedDateTime getDate() {
@@ -68,6 +91,14 @@ public class InviteEntity extends AbstractEntity {
 
     public void setDate(ZonedDateTime date) {
         this.date = date;
+    }
+
+    public void setExpiryDate(ZonedDateTime expiryDate) {
+        this.expiryDate = expiryDate;
+    }
+
+    public ZonedDateTime getExpiryDate() {
+        return expiryDate;
     }
 
     public String getCode() {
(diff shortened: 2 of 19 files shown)


src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -131,7 +134,21 @@ public void generateOtp(InviteOtpRequest inviteOtpRequest) {
...
         } else {
-            throw notFoundException();
+            throw notFoundInviteException(inviteOtpRequest.getCode());
+        }
+    }
+
+    @Transactional
+    public User createInvitedUser(InviteValidateOtpRequest inviteValidateOtpRequest) {
+        Optional<InviteEntity> inviteOptional = inviteDao.findByCode(inviteValidateOtpRequest.getCode());
+        if (inviteOptional.isPresent() && secondFactorAuthenticator.authorize(inviteOptional.get().getOtpKey(), inviteValidateOtpRequest.getOtpCode())) {
```
kakumara: How you going identify between if the invite was not found or the second factor authentication error? As discussed selfservice needs to know the differentiation in order to direct user either to re-enter otp or stop registration altogether.

kakumara: How you going identify between if the invite was not found or the second factor authentication error? As discussed selfservice needs to know the differentiation in order to direct user either to re-enter otp or stop registration altogether.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
```diff
@@ -2,58 +2,218 @@
...
                 .build();
 
         givenSetup()
                 .when()
                 .body(mapper.writeValueAsString(invitationRequest))
                 .contentType(ContentType.JSON)
-                .post(format(INVITES_OTP_RESOURCE_URL, "not-existing-code"))
+                .post(INVITES_GENERATE_OTP_RESOURCE_URL)
+                .then()
+                .statusCode(BAD_REQUEST.getStatusCode());
```
kakumara: Not exist returns BAD_REQUEST?  shouldn't it be NOT_FOUND ?

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
```diff
@@ -2,58 +2,218 @@
...
+
+        // check if the user has been created and it is not disabled in the database
+        List<Map<String, Object>> users = databaseHelper.findUserByUsername(EMAIL);
+
+        assertThat(users.size(), is(1));
+
+        Map<String, Object> createdUser = users.get(0);
+        assertThat(createdUser.get("username"), is(EMAIL));
+        assertThat(createdUser.get("email"), is(EMAIL));
+        assertThat(createdUser.get("password"), is(PASSWORD));
```
kakumara: Saving plain test password ???? this should be encrypted using encryptor.. otherwise login will not work. Have a look at UserService.java

kakumara: Instead of creating a UserEntity directly from InviteEntity .. i would rather create a User.java from the invite and pass it to UserService to save (so that it could do the relevant validations and encryptions there)

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
```diff
@@ -2,58 +2,218 @@
...
+                .put("otp", PASSCODE)
+                .build();
+
+        givenSetup()
+                .when()
+                .body(mapper.writeValueAsString(invitationOtpRequest))
+                .contentType(ContentType.JSON)
+                .post(INVITES_VALIDATE_OTP_RESOURCE_URL)
+                .then()
+                .statusCode(BAD_REQUEST.getStatusCode());
```
kakumara: Once again non-existent code becomes BAD_REQUEST? shouldn't it be NOT_FOUND?

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
```diff
@@ -2,58 +2,218 @@
...
+                .put("otp", 123456)
+                .build();
+
+        givenSetup()
+                .when()
+                .body(mapper.writeValueAsString(invitationOtpRequest))
+                .contentType(ContentType.JSON)
+                .post(INVITES_VALIDATE_OTP_RESOURCE_URL)
+                .then()
+                .statusCode(BAD_REQUEST.getStatusCode());
```
kakumara: BAD_REQUST? Shouldn't this be a 401 (UNAUTHORIZED) ?

src/main/java/uk/gov/pay/adminusers/model/User.java
```diff
@@ -18,10 +18,8 @@
     public static final String FIELD_USERNAME = "username";
     public static final String FIELD_PASSWORD = "password";
     public static final String FIELD_EMAIL = "email";
-    public static final String FIELD_SERVICE_IDS = "service_ids";
     public static final String FIELD_GATEWAY_ACCOUNT_IDS = "gateway_account_ids";
     public static final String FIELD_TELEPHONE_NUMBER = "telephone_number";
-    public static final String FIELD_OTP_KEY = "otp_key";
```
kakumara: Why are these removed from User?

src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
```diff
@@ -144,27 +175,25 @@ public Invite toInvite() {
...
-     * - Invite is not expired because 02-05-2017 17:50:10 < 03-05-2017 00:00:00
-     *
-     * @return true if expired
-     */
     public boolean isExpired() {
-        return now(UTC).isAfter(date.truncatedTo(DAYS).plus(EXPIRY_DAYS, DAYS));
+        return now(UTC).isAfter(expiryDate);
+    }
+
+    public UserEntity mapToUserEntity() {
```
kakumara: See comment below. I think it would be better to create a User.java instead of UserEntity here..

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -48,15 +48,18 @@ public InviteService(RoleDao roleDao,
                          InviteDao inviteDao,
                          AdminUsersConfig config,
                          PasswordHasher passwordHasher,
-                         NotificationService notificationService, SecondFactorAuthenticator secondFactorAuthenticator) {
+                         NotificationService notificationService,
```
kakumara: This constructor is growing and have the same symptoms to that of UserService. I would have preferred an approach similar to ServiceFactory approach (see UserServicesFactory.java) so that the business logic can scale horizontally.
Although I'm not too picky on this comment and happy to do this kind of refactoring at a later stage if its too much

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
DanailMinchev: I have added `401 - UNAUTHORIZED` with the log message `Invite for code %s provided invalid otp auth code`

kakumara:
Also it seems you are missing `@Transactional` in InviteService.generateOtp()

src/main/java/uk/gov/pay/adminusers/model/User.java
DanailMinchev: Because `FIELD_OTP_KEY ` is not used anywhere, let me know if you want me to add it again.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
DanailMinchev: Thanks for spotting that!

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
DanailMinchev: As discussed this will be part of refactoring story

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
DanailMinchev: Actually, the InviteEntity is already being used with the hashed password when it is created: https://github.com/alphagov/pay-adminusers/blob/master/src/main/java/uk/gov/pay/adminusers/service/InviteService.java#L121 and we are inserting the `PASSWORD` (`a-secure-password`) as a `InviteEntity` fixture

DanailMinchev: About the password comment:

Actually the `InviteEntity` is already being used with hashed password when it is created: https://github.com/alphagov/pay-adminusers/blob/master/src/main/java/uk/gov/pay/adminusers/service/InviteService.java#L121 and we are inserting the PASSWORD (a-secure-password) as a InviteEntity fixture, that is why we are comparing directly.

About creating the `UserEntity` directly, as discussed we will need to refactor this later, because we have only the encrypted password from `InviteEntity` and `User` or `CreateUserRequest` needs plain text version

src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
DanailMinchev: Please see my comment above..

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
kakumara: 👍

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpTest.java
kakumara: I don't think these 2 tests are really necessary (just one is enough). As discussed

PR #83, opened 2017-05-19
PP-1603 Invite invalidation

DanailMinchev:
## WHAT

Invalidate an invite when:
- creating a user from invite
- the invite is locked

## HOW

- Add `disabled` and `loginCounter` properties to `InviteEntity`
- Invalidate an `InviteEntity` when creating user from invite
- Add invite locking logic (`disabled` and `loginCounter` properties)

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Invite.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Invite.java
@@ -16,6 +16,8 @@ public class Invite {
 
     private final String email;
     private String telephoneNumber;
+    private Boolean disabled = Boolean.FALSE;
+    private Integer loginCounter = 0;
 
     private List<Link> links = new ArrayList<>();
 
@@ -23,9 +25,12 @@ public class Invite {
         this.email = email;
     }
 
-    public Invite(String email, String telephoneNumber) {
+    public Invite(String email, String telephoneNumber,
+                  Boolean disabled, Integer loginCounter) {
         this.email = email;
         this.telephoneNumber = telephoneNumber;
+        this.disabled = disabled;
+        this.loginCounter = loginCounter;
     }
 
     @JsonProperty("email")
@@ -36,6 +41,16 @@ public class Invite {
     @JsonProperty("telephone_number")
     public String getTelephoneNumber() {
         return telephoneNumber;
+    }
+
+    @JsonProperty("disabled")
+    public Boolean getDisabled() {
+        return disabled;
+    }
+
+    @JsonProperty("login_counter")
+    public Integer getLoginCounter() {
+        return loginCounter;
     }
 
     @JsonProperty("_links")
--- a/src/main/java/uk/gov/pay/adminusers/model/Invite.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Invite.java
@@ -44,7 +44,7 @@ public class Invite {
     }
 
     @JsonProperty("disabled")
-    public Boolean getDisabled() {
+    public Boolean isDisabled() {
         return disabled;
     }
 
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
@@ -55,6 +55,9 @@ public class InviteEntity extends AbstractEntity {
     @Column(name = "disabled")
     private Boolean disabled = Boolean.FALSE;
 
+    @Column(name = "login_counter")
+    private Integer loginCounter = 0;
+
     public InviteEntity() {
         //for jpa
     }
@@ -174,6 +177,14 @@ public class InviteEntity extends AbstractEntity {
 
     public void setDisabled(Boolean disabled) {
         this.disabled = disabled;
+    }
+
+    public Integer getLoginCounter() {
+        return loginCounter;
+    }
+
+    public void setLoginCounter(Integer loginCount) {
+        this.loginCounter = loginCount;
     }
 
     public Invite toInvite(String inviteUrl) {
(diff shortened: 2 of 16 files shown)


src/main/java/uk/gov/pay/adminusers/model/Invite.java
```diff
@@ -38,6 +43,16 @@ public String getTelephoneNumber() {
         return telephoneNumber;
     }
 
+    @JsonProperty("disabled")
+    public Boolean getDisabled() {
```
kakumara: shouldn't boolean properties normally named as `is`.. For e.g. `isDisabled`. (Have a look at the other examples. User.java)

src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
```diff
@@ -165,14 +171,30 @@ public void setPassword(String password) {
         this.password = password;
     }
 
+    public Boolean isDisabled() {
```
kakumara: 👍

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -90,7 +90,7 @@ public Response createUserUponOtpValidation(JsonNode payload) {
         return inviteValidator.validateOtpValidationRequest(payload)
                 .map(errors -> Response.status(BAD_REQUEST).entity(errors).build())
                 .orElseGet(() -> {
-                    User createdUser = inviteService.createInvitedUser(InviteValidateOtpRequest.from(payload));
+                    User createdUser = inviteService.createInvitedUserAndInvalidateInvite(InviteValidateOtpRequest.from(payload));
```
kakumara: Just a suggestion. While I prefer descriptive method names, my personal preference is the name to describe key intended actions rather than any side effects. So in my opinion what this service intends to do is validate the top and create the user. Disable Invite for me is a side effect rather than a key action. 
I would have named this like `validateOtpAndCreateUser`

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -141,16 +151,32 @@ public void generateOtp(InviteOtpRequest inviteOtpRequest) {
...
-                throw invalidOtpAuthCodeInviteException(inviteValidateOtpRequest.getCode());
-            }
             InviteEntity inviteEntity = inviteOptional.get();
+            if (!secondFactorAuthenticator.authorize(inviteEntity.getOtpKey(), inviteValidateOtpRequest.getOtpCode())) {
+                // "failed login attempt" logic
+                inviteEntity.setLoginCounter(inviteEntity.getLoginCounter() + 1);
+                inviteEntity.setDisabled(inviteEntity.getLoginCounter() >= loginAttemptCap);
+                inviteDao.merge(inviteEntity);
+                // throw UNAUTHORIZED / 401 exception if the otp validation failed
+                throw invalidOtpAuthCodeInviteException(inviteEntity.getCode());
```
kakumara: Are you sure this is working? You have to be very careful when throwing exceptions from a `@Transactional` method. In this case I can almost guarantee the counter logic will not work even if you call `merge` method because you are throwing an exception. I expect the transaction to rollback here and not update the counter. Have a look

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -141,16 +151,32 @@ public void generateOtp(InviteOtpRequest inviteOtpRequest) {
...
+            } else {
+                // "successful login attempt" logic
+                inviteEntity.setLoginCounter(0);
+            }
+            // check if the Invite is locked and throw UNAUTHORIZED / 401 exception if it is
+            if (inviteEntity.isDisabled()) {
+                throw inviteLockedException(inviteEntity.getCode());
```
kakumara: Once again, you should only throw exceptions from a `Transactional` methods if you know the rollback state is correct (basically you should know what you doing)

src/main/resources/migrations/00022_alter_table_invites_add_column_disabled.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter-table-invites-add-column-disabled
+ALTER TABLE invites ADD COLUMN disabled BOOLEAN NOT NULL DEFAULT FALSE;
```
kakumara: Once again you have to be careful here making `DEFAULT FALSE`. Coz in case if you had live invites in prod DB already, this would have made all of them active. However, we are safe here coz there's no live invites.

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
kakumara: Where is the integration test for login_counter ??

src/main/resources/migrations/00022_alter_table_invites_add_column_disabled.sql
DanailMinchev: Yes, currently we don't have invites in production (or test/staging environments), that is why no migration needed.

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
DanailMinchev: As discussed with @kakumara and @rauligs we could use http://www.vavr.io/ library and http://www.vavr.io/vavr-docs/#_either functionality to fix that.
Later, we decided to create temporary class which will be removed after refactoring and the need to use "Either" behaviour.

src/main/java/uk/gov/pay/adminusers/service/ValidateOtpAndCreateUserResult.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.ws.rs.WebApplicationException;
+
+/**
+ * This class is used to wrap the successful result or wrap thrown exception in result.
+ * Similar to "Either" pattern: http://www.vavr.io/vavr-docs/#_either
+ *
+ * TODO: This class should be removed after refactoring of second factor authentication
+ */
+public class ValidateOtpAndCreateUserResult {
```
kakumara: 😠  👍

PR #87, opened 2017-05-30
PP-1615 Add service name

DanailMinchev:
## WHAT

- Add new Service API object
- Add service name to the response in User API object

with @maxcbc


src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java
```diff
@@ -15,6 +16,9 @@
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "services_seq_gen")
     private Integer id;
 
+    @Column(name = "name")
+    private String name = Service.DEFAULT_NAME_VALUE;
```
kakumara: As discussed, we need to figure out how to not have this default value here. This means we may have to fix accept/end-to-end as well

PR #88, opened 2017-06-01
PP-1665 Write log messages on failed login attempts

rauligs:
## WHAT    
- Log failed login attempts indicating user 'externalId' and 'loginCounter'
- Log failed OTP attempts indicating user 'externalId' and 'loginCounter'
- Changed some debug levels to info and when user gets locked out to Warn (since is very unlikely to happen).


rauligs:
Good poing @DanailMinchev I'll add the extra logging as part of this PR

heathd:
just needs to log something when the username is not matched as well (e.g.

Failed login attempt - userid=NotMatched

PR #92, opened 2017-06-07
PP-2147 create service invite resource

kakumara:
- Introduced `InviteServiceFactory`
- Created a abstract `InviteRequest` class, with child classes for user and service invite requests.
- Added generic email validator to `EmailValidator` to validate email inputs.
- Extracted the links config as a injectable object as it could be injected in services instead of the whole configuration object.
- Feedback link in the email template currently directs to the `support` page in our product page (frontend). We may need to change this in future.

TODO:
There are couple of edge cases which we still need to discuss with @stephenmc.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Invite.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Invite.java
@@ -20,17 +20,19 @@ public class Invite {
     private Integer loginCounter = 0;
 
     private List<Link> links = new ArrayList<>();
+    private String type;
 
     public Invite(String email) {
         this.email = email;
     }
 
     public Invite(String email, String telephoneNumber,
-                  Boolean disabled, Integer loginCounter) {
+                  Boolean disabled, Integer loginCounter, String type) {
         this.email = email;
         this.telephoneNumber = telephoneNumber;
         this.disabled = disabled;
         this.loginCounter = loginCounter;
+        this.type = type;
     }
 
     @JsonProperty("email")
@@ -61,5 +63,13 @@ public class Invite {
     public void setInviteLink(String targetUrl) {
         Link inviteLink = Link.from(Link.Rel.invite, "GET", targetUrl);
         this.links = ImmutableList.of(inviteLink);
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public void setType(String type) {
+        this.type = type;
     }
 }
--- a/src/main/java/uk/gov/pay/adminusers/model/Invite.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Invite.java
@@ -61,7 +61,7 @@ public class Invite {
 
     public void setInviteLink(String targetUrl) {
         Link inviteLink = Link.from(Link.Rel.invite, "GET", targetUrl);
-        this.links = ImmutableList.of(inviteLink);
+        this.links.add(inviteLink);
     }
 
     public String getType() {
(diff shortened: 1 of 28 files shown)


src/main/java/uk/gov/pay/adminusers/model/Invite.java
```diff
@@ -14,23 +15,23 @@
 @JsonInclude(Include.NON_NULL)
 public class Invite {
 
+    private String code;
     private final String email;
     private String telephoneNumber;
     private Boolean disabled = Boolean.FALSE;
     private Integer loginCounter = 0;
```
rauligs: For someone that see loginCounter in an Invite will get very confused. Maybe we could think about another name (no for this PR, just to discuss if you agree)

src/main/java/uk/gov/pay/adminusers/service/InviteCreator.java
```diff
@@ -0,0 +1,118 @@
...
+import static uk.gov.pay.adminusers.model.InviteType.SERVICE;
+import static uk.gov.pay.adminusers.service.AdminUsersExceptions.*;
+
+public class InviteCreator {
+
+    private static final Logger LOGGER = PayLoggerFactory.getLogger(InviteCreator.class);
+    private static final String SELFSERVICE_INVITES_PATH = "invites";
+    private static final String SELFSERVICE_LOGIN_PATH = "login";
+    private static final String SELFSERVICE_FORGOTTEN_PASSWORD_PATH = "reset-password";
+    private static final String FEEDBACK_PATH = "support.html";
```
rauligs: We should consider moving paths into config, _Adminusers_ is knowing too much about where these resources outside are living. No for this PR but something we can discuss if agreed.

src/main/resources/config/config.yaml
```diff
@@ -73,6 +75,7 @@ graphitePort: ${METRICS_PORT:-8092}
 
 links:
   selfserviceUrl: ${SELFSERVICE_URL}
+  frontendUrl: ${FRONTEND_URL}
```
rauligs: I would build all the necessary links here rather than using constants with path into the app as I said before

src/test/java/uk/gov/pay/adminusers/resources/InviteUserRequestValidatorTest.java
```diff
@@ -275,4 +278,53 @@ public void validateOtpValidationRequest_shouldError_ifOtpFieldIsMissing() throw
...
+
+    @Test(expected = WebApplicationException.class)
+    public void shouldFail_ifEmailAddressNotPublicSector() throws Exception {
+        ImmutableMap<String, String> payload = ImmutableMap.of( "email", "example@example.co.uk","telephone_number", "08000001066", "password", "super-secure-password");
+        JsonNode payloadNode = objectMapper.valueToTree(payload);
+        validator.validateCreateServiceRequest(payloadNode);
+    }
+
+    @Test
+    public void shouldFail_ifInvalidTelephoneNumber() throws Exception {
```
rauligs: I would probably create one for a number with less than 11 digits and another for a telephone number with correct number of digits but containing a letter

src/main/java/uk/gov/pay/adminusers/service/InviteCreator.java
kakumara: 👍 fair point

src/main/resources/config/config.yaml
kakumara: 👍

PR #93, opened 2017-06-09
PP-2147 Update service attribute resource

kakumara:
Currently supported operation are 

 - to update service name (`op`=`replace` and `path`=`name`)
 - to update assign new gateway account ids (`op`=`add` and `path`=`gateway_account_ids`)

Updated tests case and readme

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -434,3 +434,101 @@ Content-Type: application/json
     
 }
 ```
+
+-----------------------------------------------------------------------------------------------------------
+
+## POST /v1/api/invites/service
+
+This endpoint creates an invitation to allow self provisioning new service with Pay.
+
+### Request example
+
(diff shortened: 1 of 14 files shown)


docs/api_specification.md
```diff
@@ -433,4 +433,102 @@ Content-Type: application/json
...
+      }
+   ]
+}
+```
+
+-----------------------------------------------------------------------------------------------------------
+
+## PATCH /v1/api/services/{serviceExternalId}
+
+This endpoint creates an invitation to allow self provisioning new service with Pay.
```
SandorArpa: Shouldn't it say something like 'This endpoint updates the name of the given service'?

kakumara: 👍 copy paste error. Will fix it in the next PR

PR #95, opened 2017-06-13
PP-2147 Refactor Invitation Logic

kakumara:
This PR contains the following changes.

1. Refactored out User invitation creation to its own creator supported by InviteServicesFactory.
   (Still leaving the old resource for backward compatibility. We need to remove it after selfservice consumes the new resource)
2. Marked the old resource (and related classes/methods) as deprecated so that they can be identified and removed later on.
3. Handling service invite password
4. Handling the scenario where a user has a valid service invite but attempts another service invite creation.

with @maxcbc

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceInviteCreator.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceInviteCreator.java
@@ -21,20 +21,19 @@ import static uk.gov.pay.adminusers.app.util.RandomIdGenerator.randomUuid;
 import static uk.gov.pay.adminusers.model.InviteType.SERVICE;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.*;
 
-public class InviteCreator {
+public class ServiceInviteCreator {
 
-    private static final Logger LOGGER = PayLoggerFactory.getLogger(InviteCreator.class);
-
+    private static final Logger LOGGER = PayLoggerFactory.getLogger(ServiceInviteCreator.class);
     private final InviteDao inviteDao;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/adminusers/service/ServiceInviteCreator.java
```diff
@@ -61,35 +63,48 @@ public Invite doCreate(InviteServiceRequest inviteServiceRequest) {
...
-    private void sendUserInviteNotification(InviteEntity invite) {
-        String inviteUrl = fromUri(linksConfig.getSelfserviceInvitesUrl()).path(invite.getCode()).build().toString();
-        UserEntity sender = invite.getSender();
-        notificationService.sendInviteEmail(sender.getEmail(), invite.getEmail(), inviteUrl)
-                .thenAcceptAsync(notificationId -> LOGGER.info("invite resent successfully to user, notification id [{}]", notificationId))
+    private Invite constructInviteAndSendEmail(InviteServiceRequest inviteServiceRequest, InviteEntity inviteEntity, boolean toUpdate) {
+        String inviteUrl = format("%s/%s", linksConfig.getSelfserviceInvitesUrl(), inviteEntity.getCode());
+        inviteEntity.setTelephoneNumber(inviteServiceRequest.getTelephoneNumber());
+        inviteEntity.setPassword(passwordHasher.hash(inviteServiceRequest.getPassword()));
+        if (toUpdate) {
```
rauligs: I find this condition a bit awkward, this method is doing two different things depending on a flag. Something to consider for refactoring in next PR?

PR #99, opened 2017-06-20
PP-1608 Add pact tests for DELETE user from a service

rauligs:
## WHAT   
 - Add file for contract testing, currently will be ignored, a new issue might be needed to be raised with `pact` since it seems it doesn't support DELETE operations with request body despite the RFC seems to allow bodies on DELETE:

- See: https://stackoverflow.com/questions/14323716/restful-alternatives-to-delete-request-body


rauligs:
Close as the API will change to introduce a header representing the user performing the operation

PR #100, opened 2017-06-22
PP-1608 Allow empty services for an user

rauligs:
## WHAT
- Since users can be removed from all their services all the resources should not rely on users belonging to services.

- Services returns as empty for user not belonging to a service therefore happens the same with its permissions / role.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/User.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/User.java
@@ -141,12 +141,12 @@ public class User {
      */
     @JsonProperty("role")
     public Role getRole() {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/adminusers/model/User.java
```diff
@@ -141,12 +141,12 @@ public Integer getSessionVersion() {
      */
     @JsonProperty("role")
     public Role getRole() {
-        return roles != null ? roles.get(0) : null;
+        return !roles.isEmpty() ? roles.get(0) : null;
```
DanailMinchev: 👍

PR #102, opened 2017-06-22
PP-2167 2FA authentication against multiple timewindows

maxcbc:
- `SecondFactorAuthenticator.authorize` now checks against the current 30 time window and the previous one.
- Added tests to prove this.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/app/config/SecondFactorAuthConfiguration.java
+++ b/src/main/java/uk/gov/pay/adminusers/app/config/SecondFactorAuthConfiguration.java
@@ -1,0 +1,24 @@
+package uk.gov.pay.adminusers.app.config;
+
+import javax.validation.constraints.NotNull;
+
+public class SecondFactorAuthConfiguration {
+    @NotNull
+    private int timeWindowInSeconds;
+    
+    @NotNull
+    private int validTimeWindows;
+    
+    public int getTimeWindowInSeconds() {
+        return timeWindowInSeconds;
+    }
+
+    public int getValidTimeWindows() {
+        return validTimeWindows;
+    }
+
+    public long getTimeWindowInMillis() {
+        return timeWindowInSeconds * 1000L;
+    }
+
+}
--- a/src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
@@ -1,8 +1,12 @@
 package uk.gov.pay.adminusers.service;
 
+import com.google.inject.Inject;
 import com.warrenstrange.googleauth.GoogleAuthenticator;
+import com.warrenstrange.googleauth.GoogleAuthenticatorConfig;
 import org.slf4j.Logger;
 import uk.gov.pay.adminusers.logger.PayLoggerFactory;
+
+import java.time.Clock;
 
 import static com.google.common.io.BaseEncoding.base32;
 import static org.apache.commons.lang3.StringUtils.isBlank;
@@ -11,21 +15,23 @@ import static org.apache.commons.lang3.StringUtils.isBlank;
 public class SecondFactorAuthenticator {
 
     private static final Logger logger = PayLoggerFactory.getLogger(SecondFactorAuthenticator.class);
-
     private final GoogleAuthenticator authenticator;
+    private final Clock clock;
 
-    public SecondFactorAuthenticator() {
-        this.authenticator = new GoogleAuthenticator();
+    @Inject
+    public SecondFactorAuthenticator(GoogleAuthenticatorConfig authenticatorConfig, Clock clock) {
+        this.clock = clock;
+        this.authenticator = new GoogleAuthenticator(authenticatorConfig);
     }
 
     public int newPassCode(String secret) {
         checkNull(secret);
-        return authenticator.getTotpPassword(base32().encode(secret.getBytes()));
+        return authenticator.getTotpPassword(base32().encode(secret.getBytes()), clock.millis());
     }
 
     public boolean authorize(String secret, int passcode) {
         checkNull(secret);
-        return authenticator.authorize(base32().encode(secret.getBytes()), passcode);
+        return authenticator.authorize(base32().encode(secret.getBytes()), passcode, clock.millis());
     }
 
     private void checkNull(String secret) {
--- a/src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
@@ -1,33 +1,78 @@
 package uk.gov.pay.adminusers.service;
 
+import com.warrenstrange.googleauth.GoogleAuthenticatorConfig;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+
+import static java.time.temporal.ChronoUnit.SECONDS;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class SecondFactorAuthenticatorTest {
 
     @Rule
     public ExpectedException thrown = ExpectedException.none();
 
+    private static final String SECRET = "mysecret";
+    private static final Instant INITIAL_TIME = Instant.now();
+    private static final Duration TIME_STEP = Duration.of(30, SECONDS);
+    private static final int PAST_OR_FUTURE_WINDOWS_TO_CHECK = 4;
+    private static final int PAST_PRESENT_AND_FUTURE_WINDOWS_TO_CHECK = PAST_OR_FUTURE_WINDOWS_TO_CHECK * 2 + 1;
+    private static final GoogleAuthenticatorConfig authConfig = new GoogleAuthenticatorConfig.GoogleAuthenticatorConfigBuilder()
+            .setWindowSize(PAST_PRESENT_AND_FUTURE_WINDOWS_TO_CHECK)
+            .setTimeStepSizeInMillis(TIME_STEP.toMillis())
+            .build();
+
+
+
+    private Integer passcode;
     private SecondFactorAuthenticator secondFactorAuthenticator;
+    private Clock clock;
+
+
     @Before
     public void before() throws Exception {
-        secondFactorAuthenticator = new SecondFactorAuthenticator();
+        clock = mock(Clock.class);
+        when(clock.millis()).thenReturn(INITIAL_TIME.toEpochMilli());
+        secondFactorAuthenticator = new SecondFactorAuthenticator(authConfig, clock);
+        passcode = secondFactorAuthenticator.newPassCode(SECRET);
+        
     }
 
     @Test
     public void shouldGenerateAndValidate2FAPasscode() throws Exception {
-        String secret = "mysecret";
-        Integer password = secondFactorAuthenticator.newPassCode(secret);
-
-        assertTrue(secondFactorAuthenticator.authorize(secret, password));
-        assertFalse(secondFactorAuthenticator.authorize(secret + 1, password));
+        assertTrue(secondFactorAuthenticator.authorize(SECRET, passcode));
+        assertFalse(secondFactorAuthenticator.authorize("incorrectSecret", passcode));
     }
 
+    @Test
+    public void shouldSuccess_ifAskedToValidateImmediateLastSteps2FAPasscode() throws Exception {
+        when(clock.millis()).thenReturn(INITIAL_TIME.plus(TIME_STEP).toEpochMilli());
+
+        assertTrue(secondFactorAuthenticator.authorize(SECRET, passcode));
+    }
+
+    @Test
+    public void shouldSuccess_ifAskedToValidateAValidPastSteps2FAPasscode() throws Exception {
+        when(clock.millis()).thenReturn(INITIAL_TIME.plus(TIME_STEP.multipliedBy(PAST_OR_FUTURE_WINDOWS_TO_CHECK)).toEpochMilli());
+
+        assertTrue(secondFactorAuthenticator.authorize(SECRET, passcode));
+    }
+
+    @Test
+    public void shouldError_ifAskedToValidate2FAPasscodeOlderThanLastValidStep() throws Exception {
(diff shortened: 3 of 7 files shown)


src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
```diff
@@ -25,7 +26,16 @@ public int newPassCode(String secret) {
 
     public boolean authorize(String secret, int passcode) {
         checkNull(secret);
-        return authenticator.authorize(base32().encode(secret.getBytes()), passcode);
+        long currentTime = (System.currentTimeMillis() / 1000L);
```
markyjones: See the note about perhaps an optional time paramatar into athorize method

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +30,24 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
         assertFalse(secondFactorAuthenticator.authorize(secret + 1, password));
     }
 
+    @Test
+    public void shouldSuccess_ifAskedToValidateLastSteps2FAPasscode() throws Exception {
+        String secret = "mysecret";
+        Integer password = secondFactorAuthenticator.newPassCode(secret);
+        TimeUnit.SECONDS.sleep(30);
```
markyjones: Would it be possible to inject, mock, stub the time for this to avoid sleeps @maxcbc, even a tacky optional override into the authorisze method with the time to use rather than using `System.currentTimeMillis()`

markyjones:
@maxcbc looks like the tests failed in the build in the OTP expectations, perhaps due to the sleeping? Worth exploring whether we can inject the time somehow.

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
```diff
@@ -1,31 +1,55 @@
...
+import java.time.Clock;
+
 import static com.google.common.io.BaseEncoding.base32;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 
 public class SecondFactorAuthenticator {
 
     private static final Logger logger = PayLoggerFactory.getLogger(SecondFactorAuthenticator.class);
+    private static final int VALID_PAST_TIME_WINDOWS = 4; // Plus 1 and then times 30 for maximum validity
```
markyjones: The comment has left me confused... (probably just me!) but if not could it be clarified?

maxcbc:
After talking to Till, I have extended the amount of past codes to 4. This gives us a time range of upto 2 mins 30 secs.

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
```diff
@@ -1,31 +1,55 @@
...
-        return authenticator.authorize(base32().encode(secret.getBytes()), passcode);
+        long currentTime = clock.millis();
+        /*
+         * Validate against the current 30 second time window and the amount of past windows defined in 'VALID_PAST_TIME_WINDOWS'
+         * The authenticator already checks against windows either side of the current one,
+         * Thus we only want to check against 1 less than 'VALID_PAST_TIME_WINDOWS',
+         * As the last one will be checked automatically by the authenticator
+         */
+        for (int i = 0; i < VALID_PAST_TIME_WINDOWS; i++) {
+            long windowTime = currentTime - (30000L * i);
```
markyjones: I think hardcoding the 30000L here is a bit problematic? The 30 is actually what ever the value in config is from the GoogleAuthenticatorConfig https://github.com/wstrange/GoogleAuth/blob/master/src/main/java/com/warrenstrange/googleauth/GoogleAuthenticatorConfig.java#L36

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
maxcbc: How about:
```
// To calculate maximum validity of an OTP token in seconds: ((VALID_PAST_TIME_WINDOWS + 1) * 30)
```

I'm just keen to keep this file well commented, I think it would be a complex thing to work out how to maintain without it.

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
markyjones: @maxcbc if the logic does test the window above and below... could we not just increase that setting here https://github.com/wstrange/GoogleAuth/blob/master/src/main/java/com/warrenstrange/googleauth/GoogleAuthenticatorConfig.java#L37 from 3 to 6 which would go further?

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +37,25 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
         assertFalse(secondFactorAuthenticator.authorize(secret + 1, password));
     }
 
+    @Test
+    public void shouldSuccess_ifAskedToValidateLastSteps2FAPasscode() throws Exception {
```
markyjones: I would be tempted to do a test case and put an intermediate mils figure too 30seconds and one just below the max default of past 4 like 120 seconds

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
markyjones: Cool, I think variable name gives a good description of the purpose, I would suggest moving the `((VALID_PAST_TIME_WINDOWS + 1) * 30)` comment a bit nearer the logic that does the work?

src/main/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticator.java
```diff
@@ -1,31 +1,55 @@
...
+
+    // todo: make this more performant by checking out the minimum amount of windows needed to validate against all of windows in the range specified
     public boolean authorize(String secret, int passcode) {
         checkNull(secret);
-        return authenticator.authorize(base32().encode(secret.getBytes()), passcode);
+        long currentTime = clock.millis();
+        /*
+         * Validate against the current 30 second time window and the amount of past windows defined in 'VALID_PAST_TIME_WINDOWS'
+         * The authenticator already checks against windows either side of the current one,
+         * Thus we only want to check against 1 less than 'VALID_PAST_TIME_WINDOWS',
```
markyjones: What is the 1 in this context? 1 millisecond?

src/main/java/uk/gov/pay/adminusers/app/config/SecondFactorAuthConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+    public int getValidTimeWindows() {
+        return validTimeWindows;
+    }
+
+    public long getTimeWindowInMillis() {
+        return timeWindowInSeconds * 1000L;
+    }
+
+    public int getGoogleAuthenticatorWindowSize() {
+        return (validTimeWindows * 2) - 1;
```
alexbishop1: Took me a while to figure out what this is doing. I figured it’s something like this:

* We want to accept codes for the current window and 4 past windows (total 5)
* The way the library works is it checks the current window, _n_/2 windows in the past and _n_/2 windows in the future, where _n_ is the specified number of time windows
* Therefore, this conversion is necessary to make sure we tell the library to use enough windows to get the requisite 4 past windows and the current one (i.e. 9)

Is that correct?

If so, I don’t think it really gains us anything to have properties representing both the 5 and the 9, since the 5 figure is only ever used to calculate the nine. We should just have one property and set it to 9. Stick a comment explaining what the 9 means in practice if there’s concern the rationale may not be clear.

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -1,22 +1,37 @@
...
 
 public class SecondFactorAuthenticatorTest {
 
     @Rule
     public ExpectedException thrown = ExpectedException.none();
 
     private SecondFactorAuthenticator secondFactorAuthenticator;
+    private Clock clock;
+    private GoogleAuthenticatorConfig authConfig;
+    private long baseTime;
```
alexbishop1: Rather than a `long` of milliseconds since the epoch, I think an `Instant` would be a more appropriate type here. Additionally, a name like `initialTime` might be clearer.

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -1,22 +1,37 @@
...
     @Before
     public void before() throws Exception {
-        secondFactorAuthenticator = new SecondFactorAuthenticator();
+        baseTime = Clock.systemDefaultZone().millis();
+        authConfig = new GoogleAuthenticatorConfig.GoogleAuthenticatorConfigBuilder()
+                .setWindowSize(9)
+                .setTimeStepSizeInMillis(30000)
```
alexbishop1: I think these magic numbers should be extracted into constants:

```
    private static final Instant INITIAL_TIME = Instant.now();
    private static final Duration TIME_STEP = Duration.of(30, SECONDS);
    private static final int PAST_OR_FUTURE_WINDOWS_TO_CHECK = 4;
    private static final int PAST_PRESENT_AND_FUTURE_WINDOWS_TO_CHECK = PAST_OR_FUTURE_WINDOWS_TO_CHECK * 2 + 1;
```

These constants could then be used in the tests to remove magic numbers there.

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +43,34 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
         assertFalse(secondFactorAuthenticator.authorize(secret + 1, password));
     }
 
+    @Test
+    public void shouldSuccess_ifAskedToValidateImmediateLastSteps2FAPasscode() throws Exception {
+        String secret = "mysecret";
+        Integer password = secondFactorAuthenticator.newPassCode(secret);
```
alexbishop1: These two lines appear in almost every test. Moving them to the `@Before` method would allow the tests to focus on what’s unique each time.

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +43,34 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
...
+
+        when(clock.millis()).thenReturn(baseTime + 30000);
```
alexbishop1: If you create the constants as suggested above, you can do this like so:

```
when(clock.millis()).thenReturn(INITIAL_TIME.plus(TIME_STEP).toEpochMilli());
```

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +43,34 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
...
+
+        assertTrue(secondFactorAuthenticator.authorize(secret, password));
+    }
+
+    @Test
+    public void shouldSuccess_ifAskedToValidateAValidPastSteps2FAPasscode() throws Exception {
+        String secret = "mysecret";
+        Integer password = secondFactorAuthenticator.newPassCode(secret);
+
+        when(clock.millis()).thenReturn(baseTime + 120000);
```
alexbishop1:
```
when(clock.millis()).thenReturn(INITIAL_TIME.plus(TIME_STEP.multipliedBy(PAST_OR_FUTURE_WINDOWS_TO_CHECK)).toEpochMilli());
```

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
```diff
@@ -28,6 +43,34 @@ public void shouldGenerateAndValidate2FAPasscode() throws Exception {
...
+
+        assertTrue(secondFactorAuthenticator.authorize(secret, password));
+    }
+
+    @Test
+    public void shouldError_ifAskedToValidate2FAPasscodeOlderThanLastValidStep() throws Exception {
+        String secret = "mysecret";
+        Integer password = secondFactorAuthenticator.newPassCode(secret);
+        when(clock.millis()).thenReturn(baseTime + 151000);
```
alexbishop1:
```
when(clock.millis()).thenReturn(INITIAL_TIME.plus(TIME_STEP.multipliedBy(PAST_OR_FUTURE_WINDOWS_TO_CHECK + 1)).toEpochMilli());
```

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍 👍 👍

src/test/java/uk/gov/pay/adminusers/service/SecondFactorAuthenticatorTest.java
maxcbc: 👍 👍 👍

src/main/java/uk/gov/pay/adminusers/app/config/SecondFactorAuthConfiguration.java
maxcbc: Done 👍

PR #103, opened 2017-06-23
PP 1608 Delete user API - Introducing GovUkPay-User-Context

rauligs:
## WHAT
- Add new Header `GovUkPay-User-Context`. This represents the user performing the operation. At the moment only for one resource (_Delete user_) but eventually will be in a filter mapping URL's with specific permissions. Pending discussion.

- Add delete user API pact tests using new approach using `GovUkPay-User-Context` Header.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
@@ -135,17 +135,15 @@ public class ServiceResource {
                                           @HeaderParam(HEADER_USER_CONTEXT) String userContext) {
         LOGGER.info("Service users DELETE request - serviceId={}, userId={}", serviceId, userId);
         if (isBlank(userContext)) {
-            return Response.status(Status.UNAUTHORIZED).build();
-        } else {
-            if (userId.equals(userContext)) {
-                LOGGER.info("Failed Service users DELETE request. User and Remover cannot be the same - " +
-                        "serviceId={}, removerId={}, userId={}", serviceId, userContext, userId);
-                return Response.status(CONFLICT).build();
-            }
-            serviceServicesFactory.serviceUserRemover().remove(userId, userContext, serviceId);
-            LOGGER.info("Succeeded Service users DELETE request - serviceId={}, removerId={}, userId={}", serviceId, userContext, userId);
-            return Response.status(NO_CONTENT).build();
+            return Response.status(Status.FORBIDDEN).build();
+        } else if (userId.equals(userContext)) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -131,23 +124,28 @@ public Response findUsersByServiceId(@PathParam("serviceId") Integer serviceId)
...
-                    }
-                    serviceServicesFactory.serviceUserRemover().remove(userId, remover, serviceId);
-                    LOGGER.info("Succeeded Service users DELETE request - serviceId={}, removerId={}, userId={}", serviceId, remover, userId);
-                    return Response.status(NO_CONTENT).build();
-                })
-                .orElseGet(() -> Response.status(Status.BAD_REQUEST).build());
+        if (isBlank(userContext)) {
+            return Response.status(Status.FORBIDDEN).build();
+        } else {
+            if (userId.equals(userContext)) {
```
maxcbc: Could this not be an `else if`. Would perhaps be clearer

src/main/java/uk/gov/pay/adminusers/service/ServiceUserRemover.java
```diff
@@ -29,13 +29,14 @@ public ServiceUserRemover(UserDao userDao, ServiceRoleDao serviceRoleDao) {
     public void remove(String userExternalId, String removerExternalId, String serviceExternalId) {
         LOGGER.info("User remove from service requested - serviceId={}, removerId={}, userId={}", serviceExternalId, removerExternalId, userExternalId);
         userDao.findByExternalId(userExternalId)
-                .flatMap(userEntity -> userEntity.getServicesRole(serviceExternalId)
+                .map(userEntity -> userEntity.getServicesRole(serviceExternalId)
```
maxcbc: Is there a way in which we can get rid of some of this nesting. Maybe assigning to optional variables and throwing if empty? My Java knowledge isn't up to thinking of a way to do it but I think if it could be done it would perhaps be clearer.

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
rauligs: 👍  Definitely!

src/main/java/uk/gov/pay/adminusers/service/ServiceUserRemover.java
rauligs: 👍  Done

PR #107, opened 2017-06-28
PP-1611 including gateway account id's in to the service

kakumara:
Given that this is already mapped and loaded in ServiceEntity, it makes sense to include these id's as part of the service domain model.

We can consider removing this dependency if we consider introducing a orchestration to a different layer at some point.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java
@@ -6,6 +6,7 @@ import uk.gov.pay.adminusers.model.Service;
 import javax.persistence.*;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
 import static uk.gov.pay.adminusers.app.util.RandomIdGenerator.randomUuid;
@@ -76,12 +77,15 @@ public class ServiceEntity {
         return invites;
     }
 
-    public void addGatewayAccountIds(String ... gatewayAccountIds) {
+    public void addGatewayAccountIds(String... gatewayAccountIds) {
         populateGatewayAccountIds(asList(gatewayAccountIds));
     }
 
     public Service toService() {
         Service service = Service.from(id, externalId, name);
+        service.setGatewayAccountIds(gatewayAccountIds.stream()
+                .map(idEntity -> idEntity.getGatewayAccountId())
+                .collect(Collectors.toList()));
         return service;
     }
 
--- a/src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
@@ -1,17 +1,13 @@
 package uk.gov.pay.adminusers.resources;
 
 import com.google.common.collect.ImmutableMap;
-import com.jayway.restassured.response.ValidatableResponse;
 import org.junit.Test;
 import uk.gov.pay.adminusers.model.Service;
 
-import java.util.List;
 import java.util.Map;
 
 import static com.jayway.restassured.http.ContentType.JSON;
 import static java.lang.String.format;
-import static java.util.stream.Collectors.toList;
-import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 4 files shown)


src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
```diff
@@ -43,26 +39,19 @@ public void shouldSuccess_whenCreateAServiceWithValidGatewayAccounts() throws Ex
...
-                .body("external_id", notNullValue());
-
-        String serviceExternalId = validatableResponse.extract().jsonPath().getString("external_id");
-
-        List<Map<String, Object>> gatewayAccountsForService = databaseHelper.findGatewayAccountsByService(serviceExternalId);
-        List<String> gatewayAccountIds = gatewayAccountsForService.stream()
-                .map(gaEntry -> gaEntry.get("gateway_account_id").toString()).collect(toList());
+                .body("external_id", notNullValue())
+                .body("gateway_account_ids", hasSize(2))
+                .body("gateway_account_ids[0]", is("1"))
```
DanailMinchev: Is `gateway_account_ids` elements order is dictated by request? Probably it is better to have the sorting mechanism implemented in https://github.com/alphagov/pay-adminusers/blob/master/src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java#L83, for example https://github.com/alphagov/pay-adminusers/blob/master/src/main/java/uk/gov/pay/adminusers/persistence/entity/UserEntity.java#L218

src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java
```diff
@@ -76,12 +77,15 @@ public GatewayAccountIdEntity getGatewayAccountId() {
...
 
-    public void addGatewayAccountIds(String ... gatewayAccountIds) {
+    public void addGatewayAccountIds(String... gatewayAccountIds) {
         populateGatewayAccountIds(asList(gatewayAccountIds));
     }
 
     public Service toService() {
         Service service = Service.from(id, externalId, name);
+        service.setGatewayAccountIds(gatewayAccountIds.stream()
+                .map(idEntity -> idEntity.getGatewayAccountId())
```
DanailMinchev: We can add sorting here:
```
        service.setGatewayAccountIds(gatewayAccountIds.stream()
                .map(idEntity -> idEntity.getGatewayAccountId())
                .sorted(Comparators.usingNumericComparator())
                .collect(Collectors.toList()));
```

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
```diff
@@ -43,26 +39,19 @@ public void shouldSuccess_whenCreateAServiceWithValidGatewayAccounts() throws Ex
...
-                .statusCode(201);
-
-        validatableResponse
+                .statusCode(201)
                 .body("name", is("System Generated"))
-                .body("external_id", notNullValue());
-
-        String serviceExternalId = validatableResponse.extract().jsonPath().getString("external_id");
-
-        List<Map<String, Object>> gatewayAccountsForService = databaseHelper.findGatewayAccountsByService(serviceExternalId);
```
DanailMinchev: May be it's good to check for `GatewayAccountIdEntity` entity creation?

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
kakumara: addressed. The order is not important.

src/main/java/uk/gov/pay/adminusers/persistence/entity/ServiceEntity.java
kakumara: No need of sort, as the ordering is not important

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateTest.java
kakumara: The response check now checks of the gateway_account_ids returned as part of the response. The fact that it works proves the ID's are saved in database. Before we had to get to the database to check because we never returned the saved object.

PR #110, opened 2017-07-03
PP-2032: Add database metrics to pay-adminusers

whpearson:
Registers some metrics and then pulls in the new data for the metrics whenever
the healthcheck is called.

As the pay-connector work does.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/app/healthchecks/DatabaseHealthCheck.java
+++ b/src/main/java/uk/gov/pay/adminusers/app/healthchecks/DatabaseHealthCheck.java
@@ -1,11 +1,19 @@
 package uk.gov.pay.adminusers.app.healthchecks;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheck;
 import uk.gov.pay.adminusers.app.config.AdminUsersConfig;
+import io.dropwizard.setup.Environment;
 
 import javax.inject.Inject;
 import java.sql.Connection;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/adminusers/app/healthchecks/DatabaseHealthCheck.java
```diff
@@ -36,4 +83,17 @@ protected Result check() throws Exception {
...
+        statement.execute("select * from pg_stat_database where datname='adminusers';");
+        ResultSet resultSet = statement.getResultSet();
+        resultSet.next();
+        for (String key: integerDatabaseStatsMap.keySet() ) {
+            integerDatabaseStatsMap.put(key, resultSet.getInt(key));
+        }
+        for (String key: doubleDatabaseStatsMap.keySet() ) {
+            doubleDatabaseStatsMap.put(key, resultSet.getDouble(key));
+        }
+    }
```
kakumara: Its normally a good practice to close statements and resultsets after use. 
```
resultSet.close();
statement.close();
```

kakumara:
Other than that 👍

src/main/java/uk/gov/pay/adminusers/app/healthchecks/DatabaseHealthCheck.java
whpearson: Will do.

whpearson:
I've done a different change that has the same effect. The try ( foo=bar) { } syntax closes things that are auto-closeable at the end of it.

maxcbc:
👍 cool then merge away :)

PR #113, opened 2017-07-07
PP-1615 Add complete invite resource

mrlumbu:
In order to cleanly use invites to create entities (Users, Services),
it makes sense to have an action on the Invite resource that triggers the
transformation of the Invite into the required resources.

This commit adds an endpoint of form `/v1/api/invites/{inviteCode}/complete`

which finds the specified invite, and creates user and service (if needed) based on
data in the invite. It then disables the invite.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
@@ -1,0 +1,66 @@
+package uk.gov.pay.adminusers.service;
+
+import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
+import uk.gov.pay.adminusers.model.Invite;
+import uk.gov.pay.adminusers.model.Service;
+import uk.gov.pay.adminusers.persistence.dao.InviteDao;
+import uk.gov.pay.adminusers.persistence.dao.ServiceDao;
+import uk.gov.pay.adminusers.persistence.dao.UserDao;
+import uk.gov.pay.adminusers.persistence.entity.ServiceEntity;
+import uk.gov.pay.adminusers.persistence.entity.UserEntity;
+
+import java.util.Optional;
+
+import static uk.gov.pay.adminusers.service.AdminUsersExceptions.conflictingEmail;
+import static uk.gov.pay.adminusers.service.AdminUsersExceptions.inviteLockedException;
+
+public class InviteCompleter {
+    @Inject
+    private final InviteDao inviteDao;
+    private final UserDao userDao;
+    private final ServiceDao serviceDao;
+    private final LinksBuilder linksBuilder;
+
+    @Inject
+    public InviteCompleter(InviteDao inviteDao, UserDao userDao, ServiceDao serviceDao,  LinksBuilder linksBuilder) {
+        this.inviteDao = inviteDao;
+        this.userDao = userDao;
+        this.serviceDao = serviceDao;
+        this.linksBuilder = linksBuilder;
+    }
+
+    /**
+     * Completes an invite.
+     * ie. it creates and persists a user from an invite
+     * and if it is a service invite also creates a default service.
+     * It then disables the invite.
+     */
+    @Transactional
+    public Optional<Invite> complete(String inviteCode) {
+        return inviteDao.findByCode(inviteCode)
+                .map(inviteEntity -> {
+                    if (inviteEntity.isExpired() || inviteEntity.isDisabled()) {
+                        throw inviteLockedException(inviteEntity.getCode());
+                    }
+
+                    if (userDao.findByEmail(inviteEntity.getEmail()).isPresent()) {
+                        throw conflictingEmail(inviteEntity.getEmail());
+                    }
+
+                    if (inviteEntity.isServiceType()) {
+                        ServiceEntity serviceEntity = ServiceEntity.from(Service.from());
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
```diff
@@ -0,0 +1,66 @@
...
+                    }
+
+                    if (userDao.findByEmail(inviteEntity.getEmail()).isPresent()) {
+                        throw conflictingEmail(inviteEntity.getEmail());
+                    }
+
+                    if (inviteEntity.isServiceType()) {
+                        ServiceEntity serviceEntity = ServiceEntity.from(Service.from());
+                        serviceDao.persist(serviceEntity);
+                        inviteEntity.setService(serviceEntity);
```
kakumara: Does the inviteEntity need to know about the serviceEntity here? I thought only the userEntity needs to know about it

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
```diff
@@ -0,0 +1,66 @@
...
+                    }
+
+                    UserEntity userEntity = inviteEntity.mapToUserEntity();
```
kakumara: I think you need to add a ServiceRoleEntity here. (the serviceEntity above and a 'Admin' roleEntity)

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceCompleteTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        givenSetup()
+                .when()
+                .post(INVITES_RESOURCE_URL + "/" + inviteCode + "/complete")
+                .then()
+                .statusCode(OK.getStatusCode())
+                .body("_links", hasSize(1))
+                .body("_links[0].href", matchesPattern("^http://localhost:8080/v1/api/users/[0-9a-z]{32}$"))
+                .body("_links[0].rel", is("user"))
+                .body("disabled", is(true));
```
kakumara: how about asserting the user also (by following up the user URL) ? or at least check User  created using DatabaseTestHelper ?

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
kakumara: presumably I would pass this to a UserService (or UserCreator) so that we can encapsulate user creation logic in once place.  But we can wait for that for a later day.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceCompleteTest.java
mrlumbu: yeah was thinking that. then got lazy

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
mrlumbu: think the inviteEntity.mapToUserEntity takes care of that no?

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
mrlumbu: yep good plan

mrlumbu: am leaving this here for now as I don't think it is too ugly

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
mrlumbu: I have added tests assertions below to show that the role and service are correctly set for the user (taken from the InviteEntity)

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceCompleteTest.java
mrlumbu: aded

src/main/java/uk/gov/pay/adminusers/service/InviteCompleter.java
kakumara: My feeling is that these association(s) are not good. As it is InviteEntity has 2 hard associations to UserEntity and ServiceEntity at the moment. 
My feeling is that we should;
 - treat `Invite` as a completely independent entity not having any hard associations with any other adminuser concepts. 
- Its `OK` to have a 2 invite types  (service/user) but its only to determine what this invite is for but should not know whats the final Service or User its creating.
- "service_id" and "sender_id" should be plain Strings (rather than entities) and they should only be used for
  - "service_id" of the existing service for which this new User is invited to. 
  - "sender_id" the userId of the user who invites the new member. 

Its still okay to have a `mapToUserEntity` and/or `mapToServiceEntity` purely to construct a new User or a ServiceEntity out of this invite, but they are merely utilities for copying stuff rather than modifying the invite entity itself. Any other assignments (assigning new service to a user etc.) must be done in a controller logic area.
Happy to discuss it.

kakumara: After thinking about this a bit more, I'm okay to change my mind a bit about this. To have the mapped entities, but making sure to treat them as optional everywhere (null checks, etc.)

kakumara:
LGTM

PR #121, opened 2017-07-19
PP-1615 remove deprecated otp generate endpoint

DanailMinchev:
# WHAT

- Removed deprecated `/v1/api/invites/otp/generate` endpoint. It is not used by selfservice anymore.
We use the new `/v1/api/invites/{code}/otp/generate` instead.
- Added documentation

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -674,6 +674,7 @@ Content-Type: application/json
     
 }
 ```
+
 -----------------------------------------------------------------------------------------------------------
 
 ## POST /v1/api/invites/{code}/complete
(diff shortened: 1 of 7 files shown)


docs/api_specification.md
```diff
@@ -731,3 +732,49 @@ Content-Type: application/json
...
+ "password": "a-password"
+}
+```
+
+#### Request body description (`user` invite)
+
+| Field                    | required | Description                                                      | Supported Values     |
+| ------------------------ |:--------:| ---------------------------------------------------------------- |----------------------|
+| `telephone_number`       |   X      | the phone number of the user                                     | |
+| `password`               |   X      | password for the new user                                        | |
```
kakumara: password is not mandatory

kakumara: ignore that. its mandatory 👍

PR #129, opened 2017-08-01
PP-2359 ability find a user by username

kakumara:
Currently equals to email so this can be considered as a find by email as well too.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
@@ -52,6 +52,17 @@ public class UserResource {
         this.userServicesFactory = userServicesFactory;
     }
 
+
+    @Path(USERS_RESOURCE)
+    @GET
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -52,6 +52,17 @@ public UserResource(UserServices userServices, UserRequestValidator validator, U
         this.userServicesFactory = userServicesFactory;
     }
 
+
+    @Path(USERS_RESOURCE)
+    @GET
+    @Produces(APPLICATION_JSON)
+    public Response findUser(@QueryParam("username") String username) {
+        logger.info("User FIND request - [ {} ]", username);
```
rauligs: I believe we don't want to be logging emails. Otherwise all good

kakumara: @rauligs 👍  addressed

PR #131, opened 2017-08-03
PP-2453 support service custom branding

kakumara:
The idea is to support custom branding by using a `overridable` custom css class.

The `custom_branding` field in `services` table will store the CSS as a plain text. This is not validated (not now, purposely) to support slightly different customisations per service. Perhaps in future we could consider a template.

- `PATCH v1/api/services/{externalServiceId}` resource extends to support update custom_branding
- Added `GET  v1/api/services/{externalServiceId}` and `GET v1/api/services?gatewayAccountid={id}` resources to support front end to locate/load service.
- refactored `ServiceUpdater.java` support flexible updates.


src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateCustomBrandingTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    public void shouldReplaceWithEmpty_whenUpdatingCustomBranding_withEmptyValue() throws Exception {
+        String serviceExternalId = randomUuid();
+        Service service = Service.from(randomInt(), serviceExternalId, "existing-name");
+        service.setCustomBranding("custom branding");
+        databaseHelper.addService(service, randomInt().toString());
+
+        Map<String, String> payload = ImmutableMap.of("path", "custom_branding", "op", "replace", "value", "");
```
DanailMinchev: Is it ok to include `whitespace` test as well? Like some spaces `     `

kakumara: There will not be a situation for whitespace, given that this field (at least now) will be managed by our support. They will provide the branding and we will populate the CSS for them. so in frontend we can check if this field is defined or empty.

PR #135, opened 2017-09-07
PP-2575 Retrieve user profiles from backend

ebiere:
What
====
This exposes a new API call for retrieving a subset of User details (as UserProfile) for any given list of external user IDs.
An example is if a client calls the API with 4 different and existing user external IDs, it responds with the 4 user profiles for the iDs

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/GetUserProfilesRequest.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/GetUserProfilesRequest.java
@@ -1,0 +1,31 @@
+package uk.gov.pay.adminusers.model;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class GetUserProfilesRequest {
+
+    public static final String FIELD_EXTERNAL_IDS = "external_ids";
+
+    private List<String> externalIds = newArrayList();
+
+    public GetUserProfilesRequest() {
+    }
+
+    public GetUserProfilesRequest(List<String> externalIds) {
+        this.externalIds = externalIds;
+    }
+
+    public List<String> getExternalIds() {
+        return externalIds;
+    }
+
+    public void setExternalIds(List<String> externalIds) {
+        this.externalIds = externalIds;
+    }
+}
--- a/src/main/java/uk/gov/pay/adminusers/model/UserEmail.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/UserEmail.java
@@ -1,0 +1,46 @@
+package uk.gov.pay.adminusers.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class UserEmail {
+
+
+    public static final String FIELD_EXTERNAL_ID = "external_id";
+    public static final String FIELD_EMAIL = "email";
+
+    private String externalId;
+    private String email;
+
+
+    public UserEmail(@JsonProperty("external_id") String externalId, @JsonProperty("email")  String email) {
+        this.externalId = externalId;
+        this.email = email;
+    }
+
+    public String getExternalId() {
+        return externalId;
+    }
+
+    public void setExternalId(String externalId) {
+        this.externalId = externalId;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+
+    @Override
+    public String toString() {
+        return "UserEmail{" +
+                "externalId='" + externalId + '\'' +
+                ", email='" + email + '\'' +
+                '}';
+    }
+}
--- a/src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
@@ -2,6 +2,7 @@ package uk.gov.pay.adminusers.persistence.dao;
 
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
+import uk.gov.pay.adminusers.model.UserEmail;
 import uk.gov.pay.adminusers.persistence.entity.ServiceRoleEntity;
 import uk.gov.pay.adminusers.persistence.entity.UserEntity;
 
@@ -61,4 +62,16 @@ public class UserDao extends JpaDao<UserEntity> {
                 .map(ServiceRoleEntity::getUser)
                 .collect(Collectors.toList());
     }
+
+    public List<UserEmail> findByExternalIds(List<String> externalIds) {
+
+        String queryStr = "SELECT NEW uk.gov.pay.adminusers.model.UserEmail(u.externalId, u.email) FROM UserEntity u "+
+                           "WHERE LOWER(u.externalId) in :externalIds";
+
+        return entityManager.get()
+                .createQuery(queryStr, UserEmail.class)
+                .setParameter("externalIds", externalIds.stream().map(String::toLowerCase).collect(Collectors.toList()))
+                .getResultList();
+    }
+
 }
--- a/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
@@ -7,18 +7,17 @@ import com.google.inject.Inject;
 import io.dropwizard.jersey.PATCH;
 import org.slf4j.Logger;
 import uk.gov.pay.adminusers.logger.PayLoggerFactory;
-import uk.gov.pay.adminusers.model.CreateUserRequest;
-import uk.gov.pay.adminusers.model.PatchRequest;
-import uk.gov.pay.adminusers.model.User;
+import uk.gov.pay.adminusers.model.*;
 import uk.gov.pay.adminusers.service.UserServices;
 import uk.gov.pay.adminusers.service.UserServicesFactory;
 
 import javax.ws.rs.*;
 import javax.ws.rs.core.Response;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
+import java.util.stream.Collectors;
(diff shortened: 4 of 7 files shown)


src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +79,29 @@ public Response getUser(@PathParam("externalId") String externalId) {
...
+        if (payload != null && payload.get(GetUserProfilesRequest.FIELD_EXTERNAL_IDS) != null) {
+            logger.info("Get UserProfiles request - [ {} ]", payload.toString());
+
+            List<String> externalIds = newArrayList(payload.get(GetUserProfilesRequest.FIELD_EXTERNAL_IDS).elements())
+                    .stream().map(externalIdNode -> externalIdNode.textValue()).collect(Collectors.toList());
+
+            List<UserProfile> userProfiles = userServices.findUserByExternalIds(externalIds);
+
+            if (!userProfiles.isEmpty()) {
+                return Response.status(OK).type(APPLICATION_JSON).entity(Collections.unmodifiableMap(Stream
```
rauligs: I would leave all the logic to the UserServices. This resource would parse the externalIds and response back whatever the service returns. I would not return a not found at all because a GET is semantic through its Request Uri is malformed this should return always a 200 (when the payload is correct) with empty list if no users found or populated otherwise. 400 for invalid payload. 

How many ids are going to be requested? why not a get request with a list of ids rather than a body?

src/main/java/uk/gov/pay/adminusers/model/GetUserProfilesRequest.java
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class GetUserProfilesRequest {
+
+    public static final String FIELD_EXTERNAL_IDS = "external_ids";
```
alexbishop1: Unless I’m missing something, this constant seems to be the only thing in this class that’s actually used anywhere. Do we actually need this class?

src/main/java/uk/gov/pay/adminusers/model/UserEmail.java
```diff
@@ -0,0 +1,46 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class UserEmail {
+
+
+    public static final String FIELD_EXTERNAL_ID = "external_id";
+    public static final String FIELD_EMAIL = "email";
```
alexbishop1: Neither of these fields seem to ever be used

src/main/java/uk/gov/pay/adminusers/persistence/dao/UserDao.java
```diff
@@ -61,4 +62,16 @@ public UserDao(Provider<EntityManager> entityManager) {
                 .map(ServiceRoleEntity::getUser)
                 .collect(Collectors.toList());
     }
+
+    public List<UserEmail> findByExternalIds(List<String> externalIds) {
+
+        String queryStr = "SELECT NEW uk.gov.pay.adminusers.model.UserEmail(u.externalId, u.email) FROM UserEntity u "+
```
alexbishop1: `query` would be consistent with the rest of this class

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +80,30 @@ public Response getUser(@PathParam("externalId") String externalId) {
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
 
+    @Path(USER_EMAILS_RESOURCE)
+    @POST
```
alexbishop1: Since the input to this is so simple, I wonder if it might be more straightforward to make this a `GET` and have path parameter containing a comma-separated list of external IDs. This would remove the need to check if there is a body, check that the body has a an appropriately-named property etc. It would also make using the resource a bit more ergonomic (no need to construct a body).

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +80,30 @@ public Response getUser(@PathParam("externalId") String externalId) {
...
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Response getUserEmails(JsonNode payload) {
+        if (payload != null && payload.get(GetUserProfilesRequest.FIELD_EXTERNAL_IDS) != null) {
+            logger.info("Get UserProfiles request - [ {} ]", payload.toString());
```
alexbishop1: `UserProfiles` probably wants renaming to `UserEmails` here

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +80,30 @@ public Response getUser(@PathParam("externalId") String externalId) {
...
+
+            List<String> externalIds = newArrayList(payload.get(GetUserProfilesRequest.FIELD_EXTERNAL_IDS).elements())
+                    .stream().map(externalIdNode -> externalIdNode.textValue()).collect(Collectors.toList());
```
alexbishop1: Rather than creating a temporary `ArrayList` and streaming that, you could use Guava’s `Streams`:

```
Streams.stream(payload.get(GetUserProfilesRequest.FIELD_EXTERNAL_IDS)).map(JsonNode::textValue).collect(Collectors.toList());
```

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +80,30 @@ public Response getUser(@PathParam("externalId") String externalId) {
...
+
+            List<UserEmail> userEmails = userServices.findUserByExternalIds(externalIds);
+
+            return Response.status(OK).type(APPLICATION_JSON).entity(
+                            Collections.unmodifiableMap(Stream
+                                       .of(new AbstractMap.SimpleEntry<String, List<UserEmail>>("results", userEmails))
```
alexbishop1: Can just do `new AbstractMap<>` and let the generics be inferred

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -79,6 +80,30 @@ public Response getUser(@PathParam("externalId") String externalId) {
...
+                                       .collect(Collectors.toMap((e) -> e.getKey(), (e) -> e.getValue()))
```
alexbishop1: The brackets around each `e` aren’t really gaining us much and could probably be removed

src/main/java/uk/gov/pay/adminusers/service/UserServices.java
```diff
@@ -104,6 +106,23 @@ public UserServices(UserDao userDao,
     }
 
     /**
+     * finds users by externalIds
+     *
+     * @param externalIds
+     * @return {@link User} as an {@link Optional} if found. Otherwise Optional.empty() will be returned.
+     */
```
alexbishop1: Orphaned Javadoc block?

src/main/java/uk/gov/pay/adminusers/service/UserServices.java
```diff
@@ -104,6 +106,23 @@ public UserServices(UserDao userDao,
...
+
+    /**
+     * finds users by externalIds
+     *
+     * @param externalIds
+     * @return List of matching userProfiles. Otherwise empty list is returned.
```
alexbishop1: Not `userProfiles` any more

src/test/java/uk/gov/pay/adminusers/resources/IntegrationTest.java
```diff
@@ -13,6 +13,7 @@
 public class IntegrationTest {
 
     static final String USERS_RESOURCE_URL = "/v1/api/users";
+    static final String USER_EMAILS_RESOURCE_URL = USERS_RESOURCE_URL+"/emails";
```
alexbishop1: `static final String USER_EMAILS_RESOURCE_URL = "/v1/api/users/emails"` would be consistent with the rest of the file

src/test/java/uk/gov/pay/adminusers/resources/UserResourceCreateAndGetTest.java
```diff
@@ -249,6 +251,42 @@ public void shouldReturn404_whenGetUser_withNonExistentUsername() throws Excepti
     }
 
     @Test
+    public void shouldReturn200_IfExternalIdsDoExist() throws Exception {
+        String gatewayAccount1 = valueOf(nextInt());
+        String gatewayAccount2 = valueOf(nextInt());
+        Service service = serviceDbFixture(databaseHelper).withGatewayAccountIds(gatewayAccount1, gatewayAccount2).insertService();
+        String serviceExternalId = service.getExternalId();
```
alexbishop1: This variable is unused

src/test/java/uk/gov/pay/adminusers/resources/UserResourceCreateAndGetTest.java
```diff
@@ -249,6 +251,42 @@ public void shouldReturn404_whenGetUser_withNonExistentUsername() throws Excepti
...
+        Role role = roleDbFixture(databaseHelper).insertRole();
+        User user = userDbFixture(databaseHelper).withServiceRole(service.getId(), role.getId()).insertUser();
```
alexbishop1: Cal this `user1` to match `user2` and `user3`

src/test/java/uk/gov/pay/adminusers/resources/UserResourceCreateAndGetTest.java
```diff
@@ -249,6 +251,42 @@ public void shouldReturn404_whenGetUser_withNonExistentUsername() throws Excepti
...
+                .contentType(JSON)
+                .accept(JSON)
+                .post(USER_EMAILS_RESOURCE_URL)
+                .then()
+                .statusCode(200)
+                .body("results.size()", is(3))
+                .body("results[0].email", notNullValue())
+                .body("results[0].external_id", notNullValue())
+                .body("results[2].email", notNullValue())
+                .body("results[2].external_id", notNullValue());
```
alexbishop1: I’d prefer if this were a bit more rigourous (i.e. actually check the external IDs and email are correct).

Also, no checks of `results[1]`!

alexbishop1:
Closing in favour of #139

PR #142, opened 2017-10-27
PP-2688 Merchant details related changes

DanailMinchev:
## WHAT

This PR adds "Merchant details" related changes:

- We added merchant details in `services` table [
6a37cd5]
- We added permissions to `read` and `update` the merchant details [
d5b5366]
- We added a new `PUT /v1/api/services/{serviceExternalId}/merchant-details` endpoint to edit the merchant details [affe4e3]

see commit messages for more information.

with @simad

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
+++ b/src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
@@ -1,0 +1,9 @@
+package uk.gov.pay.adminusers.exception;
+
+public class ServiceNotFoundException extends Throwable {
+
+    public ServiceNotFoundException(String serviceExternalId) {
+        super("Service with serviceExternalId = \"" + serviceExternalId + "\" NOT FOUND");
+    }
+
+}
--- a/src/main/java/uk/gov/pay/adminusers/exception/ValidationException.java
+++ b/src/main/java/uk/gov/pay/adminusers/exception/ValidationException.java
@@ -1,0 +1,17 @@
+package uk.gov.pay.adminusers.exception;
+
+import uk.gov.pay.adminusers.utils.Errors;
+
+public class ValidationException extends Throwable {
+
+    private Errors errors;
+
+    public ValidationException(Errors errors) {
+        this.errors = errors;
+    }
+
+    public Errors getErrors() {
+        return errors;
+    }
+
+}
--- a/src/main/java/uk/gov/pay/adminusers/exception/ValidationExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/adminusers/exception/ValidationExceptionMapper.java
@@ -1,0 +1,19 @@
+package uk.gov.pay.adminusers.exception;
+
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+
+public class ValidationExceptionMapper implements ExceptionMapper<ValidationException> {
+
+    @Override
+    public Response toResponse(ValidationException exception) {
+        return Response.status(BAD_REQUEST)
+                .entity(exception.getErrors())
+                .type(MediaType.APPLICATION_JSON)
+                .build();
+    }
+
+}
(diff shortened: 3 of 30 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
```diff
@@ -89,6 +93,16 @@ private boolean nonNumericGatewayAccountIds(JsonNode gatewayAccountNode) {
         return Optional.empty();
     }
 
+    public Optional<Errors> validateUpdateMerchantDetailsRequest(JsonNode payload) {
+        Optional<List<String>> missingMandatoryFields = requestValidations.checkIfExists(payload,
```
rauligs: I needed to check what does the method _checkIfExists_ because it does not only check the presence but the value is not blank. So maybe to consider renaming the method in future PR or creating a _RequestValidations_ test that reflects the intentions. 

Not for this PR but something we can consider to do later.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateMerchantDetailsTest.java
```diff
@@ -0,0 +1,105 @@
...
+                .body("merchant_details.name", is("somename"))
+                .body("merchant_details.address_line1", is("line1"))
+                .body("merchant_details.address_line2", is("line2"))
+                .body("merchant_details.address_city", is("city"))
+                .body("merchant_details.address_country", is("country"))
+                .body("merchant_details.address_postcode", is("postcode"));
+    }
+
+    @Test
+    public void shouldSuccess_whenUpdatingMerchantDetails_withoutOptionalFields() throws Exception {
```
rauligs: Could you have a chat with @heathd about this? I think we discussed the possibility to rebalance the testing pyramid. So for example the sad cases being tests in an more unit test fashion. Not sure about this tho because maybe then we need to refactor more like:
- Resources exclusively deserialise the request, call the service and return a success response afterwards.
- Services handle the request object, do whatever is intended to do and throw an specific related exception and this exception is handle in a ExceptionHandler, responding according to the exception thrown.
- This way the service can be widely unit tested and the resource is only responsible to handle the happy case. So then integration tests can comply with a better balanced testing pyramid.

src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
DanailMinchev: Thanks, addressed in [7db158c]

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateMerchantDetailsTest.java
DanailMinchev: Great idea @rauligs ! Addressed that in [1c1cac1] using DropWizard `ExceptionMapper` interface

src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.pay.adminusers.exception;
+
+public class ServiceNotFoundException extends Throwable {
+
+    public ServiceNotFoundException() {
```
rauligs: Do you need the default constructor?

src/main/java/uk/gov/pay/adminusers/exception/ValidationException.java
```diff
@@ -0,0 +1,31 @@
...
+
+import uk.gov.pay.adminusers.utils.Errors;
+
+public class ValidationException extends Throwable {
+
+    private int code;
+    private Errors errors;
+
+    public ValidationException() {
+        this(400, Errors.from("Error while processing the request"));
```
rauligs: I would leave the code out and the mapper takes care of that

src/main/java/uk/gov/pay/adminusers/exception/ValidationExceptionMapper.java
```diff
@@ -0,0 +1,17 @@
...
+
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class ValidationExceptionMapper implements ExceptionMapper<ValidationException> {
+
+    @Override
+    public Response toResponse(ValidationException exception) {
+        return Response.status(exception.getCode())
```
rauligs: The exception mapper sets the 400 error code no the Exception class

PR #145, opened 2017-12-20
BAU Add all services endpoint

mrlumbu:
In order to be able to achieve more things without accessing
the database, it is useful to have this endpoint to eg. find correct
external id for a service.

Only concern might be performance, at some point, but we are way off
that being a problem at the moment.


mrlumbu:
this isn't working and I don't want to clog things up. closing for now

PR #146, opened 2017-12-28
PP-3096: Add chamber to docker startup script

tlwr:
To get this to run in ECS we require the docker container to be able to
start up the app with chamber to get the variables from the environment.

This commit adds a new codepath to run the app with chamber if the
`CHAMBER` env var is set, and it adds a chamber binary (copied from
pay-selfservice).

solo @tlwr

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,20 @@
 FROM govukpay/openjdk:8-jre-alpine
 
+ARG CHAMBER_URL=https://github.com/segmentio/chamber/releases/download/v1.9.0/chamber-v1.9.0-linux-amd64
+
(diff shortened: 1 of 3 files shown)


rhowe-gds:
See https://github.com/alphagov/pay-selfservice/pull/551 for how to not vendor chamber into the repo

Dockerfile
```diff
@@ -1,9 +1,17 @@
 FROM govukpay/openjdk:8-jre-alpine
 
-RUN apk update
-RUN apk upgrade
```
rhowe-gds: Did you want to drop this?

PR #149, opened 2018-01-15
PP-3096 ECSify Adminusers in Test

belindac:
Update Jenkins file to deploy to ECS too and explicitly do not run tests with non ECS deployment

Goes with https://github.com/alphagov/pay-infra/pull/1148
solo @belindacowey

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,7 +51,8 @@ pipeline {
         branch 'master'
       }
       steps {
-        deploy("adminusers", "test", null, true)
+        deploy("adminusers", "test", null, true, false)
+        deployEcs("adminusers", "test", null, true, true)
       }
     }
   }


Jenkinsfile
```diff
@@ -52,6 +52,7 @@ pipeline {
       }
       steps {
         deploy("adminusers", "test", null, true)
```
tlwr: This line is:
```
deploy adminusers
with the test account
without a tag
tag it after the deploy
implicitly run the tests
```

I think it should be
```
deploy adminusers
with the test account
without a tag
tag it after the deploy
explicitly do not run the tests
```

i.e.

```
deploy("adminusers", "test", null, true, false)
```

see below

Jenkinsfile
```diff
@@ -52,6 +52,7 @@ pipeline {
...
+        deployEcs("adminusers", "test", null, true, true)
```
tlwr: This line is
```
deploy adminusers on ecs
with the test aws account
without any tags
tag it after deployment
explicitly run tests
```

this is correct but both lines will run the tests twice (afaik) hence the change proposed above

PR #150, opened 2018-01-16
PP-3096: Refactor docker-startup WRT migrations

tlwr:
Like pay-connector there should be 2 docker startup scripts:
- docker-startup.sh
- docker-startup-with-db-migration.sh

This is so ECS will not run the migrations.


whpearson:
Looks good.

PR #152, opened 2018-01-22
PP-3174 Implement database migrations for ECS

belindac:
Remove docker-startup-with-db-migration and replace with docker-startup with logic

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ ADD chamber--linux-amd64 /app/
 ADD target/*.yaml /app/
 ADD target/pay-*-allinone.jar /app/
 ADD docker-startup.sh /app/docker-startup.sh
-ADD docker-startup-with-db-migration.sh /app/docker-startup-with-db-migration.sh
 ADD run-with-chamber.sh /app/run-with-chamber.sh
 
 CMD bash ./docker-startup.sh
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -23,7 +23,6 @@ ADD chamber--linux-amd64 /app/
 ADD target/*.yaml /app/
 ADD target/pay-*-allinone.jar /app/
 ADD docker-startup.sh /app/docker-startup.sh
-ADD docker-startup-with-db-migration.sh /app/docker-startup-with-db-migration.sh
 ADD run-with-chamber.sh /app/run-with-chamber.sh
 
-CMD bash ./docker-startup.sh
+CMD bash ./run-with-chamber.sh
```
tlwr: This shouldn't run with chamber by default, otherwise pay-scripts it will run with chamber which we do not want.

docker-startup.sh
```diff
@@ -1,4 +1,17 @@
...
-java -jar *-allinone.jar waitOnDependencies *.yaml && \
-java $JAVA_OPTS -jar *-allinone.jar server *.yaml
+set -eu
+RUN_MIGRATION=${RUN_MIGRATION:-false}
+RUN_APP=${RUN_APP:-true}
+
+java -jar *-allinone.jar waitOnDependencies *.yaml
+
+if [ "$RUN_MIGRATION" == "true" ]; then
+  java -jar *-allinone.jar db migrate *.yaml
```
tlwr: This is not equivalent to the file that was deleted

PR #154, opened 2018-01-25
PP-3063 Self links for service resource in adminusers is wrong

belindac:
## WHAT

The id was being built into the url instead of the external id.
Updated test with link assertions and removed the deprecated logic using numeric service ids.
Refactored endpoints and method parameters to reflect more explicitly which id is being used.

@belindac  @danailminchev

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/LinksBuilder.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/LinksBuilder.java
@@ -29,8 +29,7 @@ public class LinksBuilder {
     }
 
     public Service decorate(Service service) {
-        //TODO: fix to use the externalId, after getting rid of the primary key based GET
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/adminusers/service/LinksBuilder.java
```diff
@@ -29,8 +29,7 @@ public User decorate(User user) {
     }
 
     public Service decorate(Service service) {
-        //TODO: fix to use the externalId, after getting rid of the primary key based GET
```
simad: This comment made me search around for places using `findById` and came across a few examples like https://github.com/alphagov/pay-adminusers/blob/b95501107acdbf22625f7da73e860a4067befdbb/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java#L153

It seems we are still accounting for the possibility of having a numeric (internal) id - it would be good to clean those places up

PR #156, opened 2018-01-31
PP-3061 Sort Gateway Accounts with ad-hoc comparator

simad:
- We are introducing external account ids which are alphanumeric. In order to maintain backwards
  compatibility with numeric internal ids we can't use the Numeric Comparator any more.
- This PR introduces a new comparator that sorts the numeric account ids before the alphanumeric ids

 with @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
@@ -54,7 +54,7 @@ public class CreateUserRequest {
                 gatewayAccountIds =
                         ImmutableList.copyOf(node.get(FIELD_GATEWAY_ACCOUNT_IDS).iterator())
                                 .stream().map(JsonNode::asText)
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
```diff
@@ -54,7 +54,7 @@ public static CreateUserRequest from(JsonNode node) {
                 gatewayAccountIds =
                         ImmutableList.copyOf(node.get(FIELD_GATEWAY_ACCOUNT_IDS).iterator())
                                 .stream().map(JsonNode::asText)
-                                .sorted(Comparators.usingNumericComparator())
+                                .sorted(Comparators.compareGatewayAccounts())
```
heathd: `Comparators.numericallyThenLexicographically()`?

simad: that is the level of precision I'd expect from you, David :D

PR #163, opened 2018-03-05
PP-3381 Instrument all stages of a master build and deployment to Test

belindac:
Use new groovy scripts with metric integration to test one isolated build
to minimise disruption with any problems.

Requires https://github.com/alphagov/pay-jenkins-library/pull/33/files and https://github.com/alphagov/pay-jenkins-library/pull/37 to be merged first.

solo @belindac

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,8 @@ pipeline {
 
   environment {
     DOCKER_HOST = "unix:///var/run/docker.sock"
+    HOSTED_GRAPHITE_ACCOUNT_ID = credentials('graphite_account_id')
+    HOSTED_GRAPHITE_API_KEY = credentials('graphite_api_key')
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -22,15 +24,33 @@ pipeline {
         sh 'docker pull govukpay/postgres:9.4.4'
         sh 'mvn clean package'
       }
+      post {
+        failure {
+          metricsUtils("adminusers.maven-build.failure", 1, "new")
+        }
+        success {
+          script {
+            def long mavenBuildDuration = (System.currentTimeMillis() - currentBuild.startTimeInMillis) / 1000
```
whpearson: Is it worth abstracting this to a function. Like 
`metricsForSuccessfulBuild microservice build-type` ?

belindac: Yes it would. I think I did the metrics implementation 20 different ways with functions, without, with classes etc and spent a lot of time trying to wrestle with groovy and jenkins that I missed this when I finally got it working.

PR #169, opened 2018-03-21
PP-3265 Jenkins jobs converted to use selenium style smoke tests for all

belindac:
microservices

Use new separate split out jobs for deployment, smoke tests, tagging and
deployment event notification
Updated deployEcs arguments not to run tests and tag

solo @belindac

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -101,7 +101,36 @@ pipeline {
         branch 'master'
       }
       steps {
-        deployEcs("adminusers", "test", null, true, true)
+        deployEcs("adminusers", "test", null, false, false)
+      }
+    }
+    stage('Smoke Tests') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -101,7 +101,36 @@ pipeline {
...
-        deployEcs("adminusers", "test", null, true, true)
+        deployEcs("adminusers", "test", null, false, false)
+      }
+    }
+    stage('Smoke Tests') {
+      when {
+        branch 'master'
+      }
+      steps {
+        runCardSmokeTest()
```
mrlumbu: card smoke test doesn't really hit adminusers
I would specify the products smoke test for now, but we probably need some sort of very light selfservice login smoke test eventually

PR #176, opened 2018-04-03
PP-3576 Add new 2FA properties to user

alexbishop1:
Add properties for second factor, provisional OTP key and provisional OTP key created at timestamp, to match the new database columns in the users table.

When creating a new user, default the second factor to be SMS and have the other two properties be null.


src/main/java/uk/gov/pay/adminusers/model/User.java
```diff
@@ -185,14 +224,15 @@ public void setLinks(List<Link> links) {
     }
 
     /**
-     * its probably not a good idea to toString() password / otpKey
+     * it’s definitely not a good idea to toString() password / otpKey
```
DanailMinchev: :D 👍

PR #180, opened 2018-04-12
PP-3635 Whitelist National History Museum on account creation

rauligs:
## WHAT
    
- Whitelisting all '@nhm.ac.uk' addresses to pass the email validation.


src/test/java/uk/gov/pay/adminusers/utils/email/EmailValidatorIsPublicSectorEmailTest.java
```diff
@@ -78,6 +78,7 @@ public EmailValidatorIsPublicSectorEmailTest(String email, boolean testResult) {
                 {"valid@slc.co.uk", true},
                 {"valid@ucds.email", true},
                 {"valid@acas.org.uk", true},
+                {"valid@acas.nhm.ac.uk", true},
```
DanailMinchev: I think here we are testing the "main domains" in the list, for example `valid@nhm.ac.uk`, because we have "subdomains" section below

rauligs: ups... copy pasting exception. Thanks!

PR #190, opened 2018-04-24
PP-3669 Add endpoint for sending direct debit emails

simad:
- Adminusers is now in charge of sending emails on behalf of direct debit connector.
   This PR adds and endpoint so that direct debit connector can send adminusers details about
   the email that needs to be sent.
 - Non direct debit specific information (like service name or merchant details) are already
   stored in Adminusers, so we need to "merge" the two personalisations to send emails.
 - Adding direct debit notify configurations (template ids and a different api keys)

with @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/EmailRequestParser.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/EmailRequestParser.java
@@ -22,7 +22,7 @@ public class EmailRequestParser {
     EmailRequest parse(JsonNode payload) throws InvalidEmailRequestException {
         try {
             String emailAddress = payload.get("address").asText();
-            String gatewayAccountId = payload.get("gateway_account_id").asText();
+            String gatewayAccountId = payload.get("gateway_account_external_id").asText();
             EmailTemplate template = EmailTemplate.fromString(payload.get("template").asText());
             Map<String, String> personalisation = mapper.convertValue(payload.get("personalisation"), Map.class);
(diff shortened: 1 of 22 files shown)


src/main/resources/config/config.yaml
```diff
@@ -60,7 +60,8 @@ proxy:
   port: ${HTTP_PROXY_PORT}
 
 notify:
-  apiKey: ${NOTIFY_API_KEY:-api_key-pay-notify-service-id-pay-notify-secret-needs-to-be-32-chars-fsghdngfhmhfkrgsfs}
+  cardApiKey: ${NOTIFY_API_KEY:-api_key-pay-notify-service-id-pay-notify-secret-needs-to-be-32-chars-fsghdngfhmhfkrgsfs}
```
DanailMinchev: Probably it's good to rename that to `NOTIFY_CARD_API_KEY` in the future

src/main/java/uk/gov/pay/adminusers/resources/EmailRequestParser.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Inject
+    public EmailRequestParser(ObjectMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    EmailRequest parse(JsonNode payload) throws InvalidEmailRequestException {
+        try {
+            String emailAddress = payload.get("address").asText();
+            String gatewayAccountId = payload.get("gateway_account_id").asText();
```
DanailMinchev: Can we pass and use `gateway_account_external_id`s instead of internal ones?

PR #201, opened 2018-05-08
bau: do not pull docker image

oswaldquek:
It's already pulled in https://github.com/alphagov/pay-adminusers/blob/e773788a06217816e970217e2e6a814899df73a3/src/test/java/uk/gov/pay/adminusers/infra/PostgresContainer.java#L73.

@oswaldquek


oswaldquek:
DO NOT MERGE

PR #204, opened 2018-05-17
PP-3712 Allow alphabetic gateway account ids when finding services

simad:
- We were assuming that gateway account ids are numeric. This has changed
   since introducing Direct Debit. We need to allow searching by external
   gateway account id, so this PR removes the validation rules which requires
   ids to be numeric.

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/resources/ServiceRequestValidatorTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/resources/ServiceRequestValidatorTest.java
@@ -25,7 +25,7 @@ public class ServiceRequestValidatorTest {
     private ServiceRequestValidator serviceRequestValidator = new ServiceRequestValidator(new RequestValidations());
 
     @Test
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/adminusers/resources/ServiceRequestValidatorTest.java
```diff
@@ -25,14 +25,20 @@
...
+    public void shouldSuccess_whenUpdate_whenAllFieldPresentAndValid() {
         ImmutableMap<String, String> payload = ImmutableMap.of("path", "name", "op", "replace", "value", "example-name");
 
         Optional<Errors> errors = serviceRequestValidator.validateUpdateAttributeRequest(mapper.valueToTree(payload));
 
         assertFalse(errors.isPresent());
     }
 
+    @Test
+    public void shouldAllowNonNumericGatewayAccounts_whenFindingServices() {
```
DanailMinchev: 👍

PR #205, opened 2018-05-17
Pp 3629 verify provider deployment

oswaldquek:
Verify the pact just before deploy. Have chosen to do this in the Deploy stage  as we already have too many stages. `checkPactCompatibility` is provider/consumer agnostic so we just in the service name, commit sha and environment to it.
See https://build.ci.pymnt.uk/job/pay-adminusers/view/change-requests/job/PR-205/1/console and search for `"deployable":true,"reason":"All verification results are published and successful"` for a working demonstration.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,6 +93,7 @@ pipeline {
         branch 'master'
       }
       steps {
+        checkPactCompatibility("adminusers", gitCommit(), "test")
         deployEcs("adminusers")
       }
     }


Jenkinsfile
```diff
@@ -95,6 +95,7 @@ pipeline {
         branch 'master'
       }
       steps {
+        checkPactCompatibility("adminusers", gitCommit(), "test")
```
mrlumbu: https://github.com/alphagov/pay-jenkins-library/blob/master/vars/checkPactCompatibility.groovy requires the `consumer` and `consumerVersion` , but you are passing in the provider. Bit confusing

mrlumbu:
Could you describe in the commit and PR description a bit more about how this works

Jenkinsfile
oswaldquek: ah. it's actually checkPactCompatibility that needs fixing. it's arguments should be `service` and `version`, not `consumerName` and `consumerVersion` as it doesn't matter whether you pass in the consumer or provider. i'll raise a PR for that.

mrlumbu:
@oswaldquek happy to merge once you have added a bit more detail to commit message and PR description

oswaldquek:
@mrlumbu commit message and PR description udpated

PR #208, opened 2018-05-21
PP-3713 Convert country iso code to human readable name

simad:
- `MerchantDetailsEntity` does not hold the name of a country but its
   iso code, which needs to be mapped to a human readable name.
   We are using the same countries file as our node microservices - this should
   really be shared between all of our apps. For now, we unfortunately have to
   copy it in adminusers, too.
 - The converter creates a map at startup, to avoid having to parse the file (whose
   format is an array of arrays instead of a simple map) multiple times.
 - The human readable country name is currently displayed on direct debit emails, as
   part of the organisation address


src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
```diff
@@ -0,0 +1,64 @@
...
+    public CountryConverter(ObjectMapper objectMapper) throws IOException {
+        this.objectMapper = objectMapper;
+        String countries = Resources.toString(Resources.getResource(COUNTRIES_FILE_PATH), Charsets.UTF_8);
+        this.countries = createMap(countries);
+    }
+
+    private Map<String, String> createMap(String countries) throws IOException {
+        List<List<String>> allCountries = objectMapper.readValue(countries, new TypeReference<List>() {});
+        return allCountries.stream()
+                .map(a -> new Country(a.get(0), a.get(1)))
```
alexbishop1: Obviously this will blow up all over the place if the JSON (and even the `country:XX` strings within the JSON) aren’t correct. I’m OK with that. The `countries.json` file isn’t going to change without our knowledge and I’d rather that the app didn’t start in that case than stumble into life and try to use some bad data file.

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
```diff
@@ -0,0 +1,64 @@
...
+                .filter(Country::isCountry)
+                .collect(Collectors.toMap(Country::getIsoCode, Country::getName)
+        );
+    }
```
alexbishop1: So forgive me, but I spent quite a while pondering if it’s possible to attain a higher state of lambda transcendence, where we don’t need the intermediary `Country` object at all.

I came up with this:

```java
private Map<String, String> createMap(String countries) throws IOException {
    List<List<String>> allCountries = objectMapper.readValue(countries, new TypeReference<List<String>>() {});
    Pattern countryColonLiteralThenCode = Pattern.compile("^country:([A-Z][A-Z])$");
    return allCountries.stream()
            .filter(country -> countryColonLiteralThenCode.matcher(country.get(1)).matches())
            .collect(toMap(country -> country.get(1).substring(country.get(1).indexOf(':')), country -> country.get(0)));
    }
```

Thoughts? I quite like that it cuts the class size by about a third but I’m not sure if it’s really much clearer.

src/test/java/uk/gov/pay/adminusers/utils/CountryConverterTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public void setUp() throws Exception {
+        this.countryConverter = new CountryConverter(
+                objectMapper);
+    }
+
+    @Test
+    public void shouldGetCountryNameFromIsoName() {
+        assertThat(countryConverter.getCountryNameFrom("AA").get(), is("aaa"));
+        assertThat(countryConverter.getCountryNameFrom("BB").get(), is("bbb"));
+        assertThat(countryConverter.getCountryNameFrom("CC").isPresent(), is(false));
```
alexbishop1: Maybe also something that isn’t in the file at all?

src/test/java/uk/gov/pay/adminusers/service/EmailServiceTest.java
```diff
@@ -182,13 +190,14 @@ public void shouldThrowAnExceptionIfMerchantDetailsAreMissing() throws InvalidMe
...
                 POSTCODE,
-                ADDRESS_COUNTRY,
+                ADDRESS_COUNTRY_CODE,
                 MERCHANT_EMAIL
         );
 
         given(mockServiceDao.findByGatewayAccountId(GATEWAY_ACCOUNT_ID)).willReturn(Optional.of(mockServiceEntity));
         given(mockServiceEntity.getMerchantDetailsEntity()).willReturn(merchantDetails);
         given(mockServiceEntity.getName()).willReturn("a service");
+        given(mockCountryConverter.getCountryNameFrom(ADDRESS_COUNTRY_CODE)).willReturn(Optional.of("Cake Land"));
```
alexbishop1: This line appears four times: can we put it in a `@Before` method?

src/main/java/uk/gov/pay/adminusers/service/EmailService.java
```diff
@@ -46,7 +48,8 @@ private String formatMerchantAddress(MerchantDetailsEntity merchantDetails) {
         }
         merchantAddress.add(merchantDetails.getAddressCity());
         merchantAddress.add(merchantDetails.getAddressPostcode());
-        merchantAddress.add(merchantDetails.getAddressCountry());
+        countryConverter.getCountryNameFrom(merchantDetails.getAddressCountryCode())
+                .ifPresent(merchantAddress::add);
```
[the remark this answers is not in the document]
simad: Yes, I have been thinking what would happen if we have an invalid iso code. This _shouldn't_ happen, unless the countries files got out of sync between frontend / adminusers (which is a risk of not having a single source of truth) or _maybe_ people purposefully tamper with our payloads, I think.

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.stream.Collectors;
+
+public class CountryConverter {
+
+    public static final String COUNTRIES_FILE_PATH = "countries.json";
+
+    private static class Country {
+        private final String name;
+        private String isoCode;
+        private String type;
```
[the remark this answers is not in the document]
simad: 👍

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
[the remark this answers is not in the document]
simad: NO

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Inject
+    public CountryConverter(ObjectMapper objectMapper) throws IOException {
+        this.objectMapper = objectMapper;
+        String countries = Resources.toString(Resources.getResource(COUNTRIES_FILE_PATH), Charsets.UTF_8);
+        this.countries = createMap(countries);
+    }
+
+    private Map<String, String> createMap(String countries) throws IOException {
+        List<List<String>> allCountries = objectMapper.readValue(countries, new TypeReference<List>() {});
```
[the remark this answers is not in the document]
simad: Almost, it's TypeReference<List<List<String>>> 💯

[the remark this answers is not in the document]
alexbishop1: Not `TypeReference<List<List<String>>>`?

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
simad: The reason why I had such a verbose way of dealing with things (inner class etc.) was indeed readability.
I like your solution, though! Trying to make it a bit more readable I came up with

```
     List<List<String>> allCountries = objectMapper.readValue(countries, new TypeReference<List<List<String>>>() {});
        return allCountries.stream()
                .filter(country -> country.get(1).startsWith("country:"))
                .collect(toMap(
                        country -> getIsoCode(country.get(1)),
                        country -> country.get(0)));
```
where `getIsoCode` does the substring bit. What do you think?

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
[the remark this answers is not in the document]
simad: yes, that one. I can't copy paste #neverforget

src/main/java/uk/gov/pay/adminusers/utils/CountryConverter.java
```diff
@@ -0,0 +1,44 @@
...
+        List<List<String>> allCountries = objectMapper.readValue(countries, new TypeReference<List<List<String>>>() {});
+        return allCountries.stream()
+                .filter(country -> country.get(1).startsWith("country:"))
+                .collect(toMap(
+                        country -> getIsoCode(country.get(1)),
+                        country -> country.get(0)));
+    }
+    
+    private static String getIsoCode(String typeAndIsoCode) {
+        return typeAndIsoCode.substring(typeAndIsoCode.indexOf(':') + 1);
```
alexbishop1: Caught my off-by-one error, nice!

src/test/java/uk/gov/pay/adminusers/utils/CountryConverterTest.java
alexbishop1: So… about that?

PR #211, opened 2018-06-08
PP-2495 AWS x-ray on application

belindac:
Adds incoming request tracing/segmenting. This allows the flowing of traces from upstream services (self-service) into connector, which can then be visualised using the aws console, or called via the API.

Outbound http requests to any downstream services is a WIP and was previously failing build/test, so we are splitting the work up in order to get the quick win/advantages above.


belindac:
We are duplicating AWS code ... did Oz not do work to make a common java module ? We should put this code there, but understand quick win now.

PR #218, opened 2018-07-02
PP-3676 Send email on mandate created

georgeracu:
- Added keys for email template to be sent when a mandate is created

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/EmailService.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/EmailService.java
@@ -87,28 +87,39 @@ public class EmailService {
                 ORGANISATION_PHONE_NUMBER_KEY, merchantDetails.getTelephoneNumber(),
                 ORGANISATION_EMAIL_ADDRESS_KEY, merchantDetails.getEmail()
         );
-        return ImmutableMap.of(
-                EmailTemplate.ONE_OFF_PAYMENT_CONFIRMED, new StaticEmailContent(
+
+        return new HashMap<EmailTemplate, StaticEmailContent>() {
+            {
+                put(EmailTemplate.ONE_OFF_PAYMENT_CONFIRMED, new StaticEmailContent(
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/adminusers/service/EmailService.java
```diff
@@ -87,28 +87,39 @@ private boolean isMissingMandatoryFields(MerchantDetailsEntity merchantDetails)
...
                         notificationService.getNotifyDirectDebitConfiguration().getMandateFailedEmailTemplateId(),
                         personalisation
-                )
-        );
+                ));
+                put(EmailTemplate.ON_DEMAND_MANDATE_CREATED, new StaticEmailContent(
+                        notificationService.getNotifyDirectDebitConfiguration().getOnDemandMandateCreatedEmailTemplateId(),
+                        personalisation
+                ));
+                put(EmailTemplate.ONE_OFF_MANDATE_CREATED, new StaticEmailContent(
```
DanailMinchev: We send one email for mandate and payment creation for `ONE_OFF` payments - https://github.com/alphagov/pay-direct-debit-connector/blob/56788680d1c42151fff951b6c8f0610067bf2019/src/main/java/uk/gov/pay/directdebit/notifications/services/UserNotificationService.java#L55

In this case you are separating them and calling the same template `getOneOffMandateAndPaymentCreatedEmailTemplateId()` so we are going to send the same email twice if we use this functionality from connector.

As discussed I think this is fine, because adminusers is prepared to handle one off mandate creation in case it is needed in the future and we just need to change `getNotifyDirectDebitConfiguration().getOnDemandMandateCreatedEmailTemplateId()` with appropriate configuration.

Just remember not to use this in connector for now and probably it's good to add a comment there.

Thanks

georgeracu: Yes, to avoid sending the same email twice, we will send it only once for payment, but skip it when we setup direct debit for ONE_OFF. This is done in direct debit connector.

PR #226, opened 2018-07-27
[DO NOT MERGE] PP-3983 Added state/method to support more specific user, service and gateway account ids

brid:
[WIP] Further details to come.


jonheslop:
This PR is redundant

PR #229, opened 2018-08-15
PP-4096 Add service_name object on service response

georgeracu:
PP-4096 Added `service_name` to the GET `service` resource

- Added new entity for ServiceName
- Added enum for supported language and converter for JPA (should be moved in commons repo)
- Updated tests
- If there's no record in the new table for the service name, then we
    populate the `service_name` object with the same value as the one from
    field `name`

PP-4096 Opportunistic refactoring

- Refactored `ServiceDaoTest` to use entity builders instead of repeating
    code to create the same entities
- Added method to `DatabaseTestHelper` to persist an entire `ServiceEntity` and
    all its related objects, instead of inserting raw values
- Extracted builder objects for `MerchantDetailsEntity` and `ServiceEntity`
- Updated test to use the new `SupportedLanguage` instead of hardcoded strings

PP-4096 Opportunistic refactoring

- Refactored `ServiceResourceFindTest` to use mocked Dao instead of using
    the database. The `ServiceDao` is tested separately and the fields are asserted
    into those tests.
- Decoupling from the database, it makes running the tests faster. A run
    with Maven command line shows a difference of 26s between the tests hitting
    the db and the ones using mocked Dao


src/main/java/uk/gov/pay/adminusers/persistence/entity/service/ServiceNameEntity.java
```diff
@@ -0,0 +1,95 @@
...
+    private SupportedLanguage language;
+
+    @Column(name = "name")
+    private String name;
+
+    public ServiceNameEntity() {
+        // make JPA happy
+    }
+
+    //region <Getters/Setters>
```
DanailMinchev: `region`s 🎉

src/main/java/uk/gov/pay/adminusers/persistence/entity/service/SupportedLanguage.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.pay.adminusers.persistence.entity.service;
+
+public enum SupportedLanguage {
```
DanailMinchev: As discussed this will be moved to common repo later

PR #230, opened 2018-08-16
PP-4096 Save `service_name` to db

georgeracu:
- Amended resource when creating a service to add the `service_name` if
exists
- Amended resource when creating a service and there is no `service_name`
specified to add a record that matches language ENGLISH and name the same
as the `name` parameter
- Amended ServiceCreatorTest for the field, added checks for links
- Updated ServiceDaoTest to test that the new entity `ServiceNameEntity`
is successfully persisted to the database
- Switched the order in which a ServiceNameEntity is associated with a
ServiceEntity such that when we call `add` on the set, then it will call
`equals()` on ServiceNameEntity and both entities will have the same ServiceEntity.
- Removed `id` from `equals` method as two ServiceNameEntity should be
considered equal when they have the same language, name and belong to the
same service.
- Replaced lambda method calls with reference method calls
- Replaced tests in ServiceCreateTest with unit tests and mocks for the service
- Removed tests from ServiceCreateTest that were calling the database, as
the whole chain is verified with tests for resource, service and dao
- Extracted common methods/fields into a base ServiceResourceBaseTest class
- Added helper method on the JDBI database test helper to get a list of
ServiceNameEntity based on serviceId


src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceBaseTest.java
```diff
@@ -0,0 +1,36 @@
...
+    static final LinksBuilder linksBuilder = new LinksBuilder(HTTPS_BASE_URL);
+
+    static void assertLinks(String serviceExternalId, JsonPath json) {
+        assertThat(json.getList("_links"), hasSize(1));
+        assertThat(json.get("_links[0].href"), is(HTTPS_BASE_URL + "/v1/api/services/" + serviceExternalId));
+        assertThat(json.get("_links[0].method"), is("GET"));
+        assertThat(json.get("_links[0].rel"), is("self"));
+    }
+
+    static void assertEnServiceNameJson(String name, JsonPath json) {
```
DanailMinchev: very minor: you can call this `enName` to be consistent with the other method in the next PR :)

src/main/java/uk/gov/pay/adminusers/persistence/entity/service/ServiceNameEntity.java
```diff
@@ -42,6 +42,14 @@ public ServiceNameEntity() {
         // make JPA happy
     }
 
+    public static ServiceNameEntity from(SupportedLanguage language, String name) {
```
DanailMinchev: As discussed you can use `SupportedLanguage` from our shared library in the next pr

PR #235, opened 2018-08-24
PP-4095 Refactor service update attribute request validation

alexbishop1:
When the update service attribute endpoint was changed to also accept an array of objects rather than just a single object, the `ServiceRequestValidator.validateUpdateAttributeRequest` method was modified to figure out if it was passed an array and validate every element in the array if so. This made the `ServiceRequestValidator` more complex. This commit simplifies things somewhat by doing the following (this is a conceptual description of a series of changes; the final result melds them all in to one):

- Change `ServiceRequestValidator.validateUpdateAttributeRequest` so it no longer tries to handle an array and assumes it just has a single object

- Move `ServiceRequestValidator.validateUpdateAttributeRequest` and its dependent private methods out in to its own `ServiceUpdateOperationValidator` class

- Change the old `ServiceRequestValidator.validateUpdateAttributeRequest` so that calls the new `ServiceUpdateOperationValidator.validate` method if it receives an object. If it receives an array, it calls that method repeatedly, once for each object in the array.

This does not (should not!) change the actual validation that’s performed but it changes the order of the checks performed on an array, which will change the errors that are returned in some cases.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
@@ -1,34 +1,19 @@
 package uk.gov.pay.adminusers.resources;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ArrayNode;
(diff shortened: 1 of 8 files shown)


alexbishop1:
Wrong branch name! Replaced by https://github.com/alphagov/pay-adminusers/pull/236.

PR #236, opened 2018-08-24
PP-4096 Refactor service update attribute request validation

alexbishop1:
When the update service attribute endpoint was changed to also accept an array of objects rather than just a single object, the `ServiceRequestValidator.validateUpdateAttributeRequest` method was modified to figure out if it was passed an array and validate every element in the array if so. This made the `ServiceRequestValidator` more complex. This commit simplifies things somewhat by doing the following (this is a conceptual description of a series of changes; the final result melds them all in to one):

- Change `ServiceRequestValidator.validateUpdateAttributeRequest` so it no longer tries to handle an array and assumes it just has a single object

- Move `ServiceRequestValidator.validateUpdateAttributeRequest` and its dependent private methods out in to its own `ServiceUpdateOperationValidator` class

- Change the old `ServiceRequestValidator.validateUpdateAttributeRequest` so that calls the new `ServiceUpdateOperationValidator.validate` method if it receives an object. If it receives an array, it calls that method repeatedly, once for each object in the array.

This does not (should not!) change the actual validation that’s performed but it changes the order of the checks performed on an array, which will change the errors that are returned in some cases.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
@@ -1,0 +1,107 @@
+package uk.gov.pay.adminusers.resources;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableMap;
+import uk.gov.pay.adminusers.validations.RequestValidations;
+
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static java.lang.String.format;
+import static uk.gov.pay.adminusers.model.ServiceUpdateRequest.FIELD_OP;
+import static uk.gov.pay.adminusers.model.ServiceUpdateRequest.FIELD_PATH;
+import static uk.gov.pay.adminusers.model.ServiceUpdateRequest.FIELD_VALUE;
+
+public class ServiceUpdateOperationValidator {
+    
+    static final String FIELD_NAME = "name";
+    static final String FIELD_GATEWAY_ACCOUNT_IDS = "gateway_account_ids";
+    static final String FIELD_CUSTOM_BRANDING = "custom_branding";
+    static final String FIELD_SERVICE_NAME = "service_name";
+    
+    private static final String REPLACE = "replace";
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class ServiceUpdateOperationValidatorTest {
+
+    private ObjectMapper mapper = new ObjectMapper();
+    private ServiceUpdateOperationValidator serviceUpdateOperationValidator = new ServiceUpdateOperationValidator(new RequestValidations());
+
+    @Test
+    public void shouldSuccess_whenUpdate_withAllFieldsPresentAndValid() {
```
danworth: I'm sure this convention and you've copied this as-is but why is it `shouldSuccess` rather than `shouldSucceed` (we use `shouldFail` instead of `shouldFailure`)?

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -0,0 +1,107 @@
...
+        if (!errors.isEmpty()) {
+            return errors;
+        }
+
+        return Collections.emptyList();
+    }
+
+    private List<String> validateOpAndPathExistAndNotEmpty(JsonNode operation) {
+        List<String> errors = new ArrayList<>();
+        requestValidations.checkIfExistsOrEmpty(operation, FIELD_OP, FIELD_PATH).ifPresent(errors::addAll);
```
danworth: Bit of a shame we need to do this due to the `requestValidations` methods return `Optional<List<String>>` rather than just `List<String>` but at least we're following a better pattern here 👍

src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
alexbishop1: Consistency with the shocking grammar used elsewhere!

PR #237, opened 2018-08-24
PP-4096 Change the API for multilingual service name updates

alexbishop1:
For updating the multilingual service name, we used to expect requests like this:

```json
[
  {
    "op": "replace",
    "path": "service_name",
    "value": {
      "en": "Rectify your renal records"
    }
  }
]
```

Now we expect requests like this:

```json
[
  {
    "op": "replace",
    "path": "service_name/en",
    "value": "Rectify your renal records"
  }
]
```

In addition to being more in line with [JSON Patch](http://jsonpatch.com/) (but note we are still not strictly compliant), this also makes it easier to change multiple language variants of the service name at once in a natural fashion:

```json
[
  {
    "op": "replace",
    "path": "service_name/en",
    "value": "Rectify your renal records"
  },
  {
    "op": "replace",
    "path": "service_name/cy",
    "value": "Addaswch eich cofnodion arennol"
  }
]
```

Furthermore, it makes it simple to reuse our existing validation for service names (no longer than 50 characters), so we do that now. We also now validate that the specified language is actually one we support.

Although this change breaks compatibility, it’s safe to do because nothing is using the other API yet (it’s an internal API, not a public one).

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.adminusers.service;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 import uk.gov.pay.adminusers.exception.ServiceNotFoundException;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
```diff
@@ -87,9 +93,12 @@ public Service doUpdateMerchantDetails(String serviceExternalId, UpdateMerchantD
         return (serviceUpdateRequest, serviceEntity) -> serviceEntity.setCustomBranding(serviceUpdateRequest.valueAsObject());
     }
 
-    private BiConsumer<ServiceUpdateRequest, ServiceEntity> updateServiceNameObject() {
+    private BiConsumer<ServiceUpdateRequest, ServiceEntity> updateMultilingualServiceName() {
```
danworth: I like this name change, makes it clearer 👍

PR #242, opened 2018-08-29
PP-4096 Update docs now we have multilingual service names

alexbishop1:
Update adminusers internal API docs to reflect that we now return and support setting translated variants of the service name.

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -567,6 +567,10 @@ Content-Type: application/json
 {
     "name": "abcd1234",
     "gateway_account_ids": ["1"],
+    "service_name": {
+      "en": "abcd1234",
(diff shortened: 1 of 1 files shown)


docs/api_specification.md
```diff
@@ -941,10 +954,13 @@ Content-Type: application/json
 
 ## PATCH /v1/api/services/`{serviceExternalId}`
 
-This endpoint modifies updatable attributes of a Service. Currently supports
+This endpoint modifies updatable attributes of a aervice. Currently supports:
```
jonheslop: aervice => service

alexbishop1: Thanks, fixed.

PR #243, opened 2018-09-03
PP-4096 Allow blank multilingual service names when updating

alexbishop1:
When receiving a request to replace one of a service’s multilingual service names, allow empty or all-whitespace strings because — in combination with the feature that means we won’t ever return a
blank service name when someone retrieves a service — this is the best way we have for users to remove a multilingual service name right now.

However, don’t allow an English service name (JSON path `service_name/en`) to be blank because that is mandatory.

with @danworth

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
@@ -77,7 +77,11 @@ public class ServiceUpdateOperationValidator {
         if (FIELD_CUSTOM_BRANDING.equals(path)) {
             errors.addAll(checkIfValidJson(operation.get(FIELD_VALUE), FIELD_CUSTOM_BRANDING));
         } else if (FIELD_NAME.equals(path) || path.startsWith(FIELD_SERVICE_NAME_PREFIX)) {
-            requestValidations.checkIfExistsOrEmpty(operation, FIELD_VALUE).ifPresent(errors::addAll);
+            if (FIELD_NAME.equals(path) || path.endsWith('/' + SupportedLanguage.ENGLISH.toString())) {
+                requestValidations.checkIfExistsOrEmpty(operation, FIELD_VALUE).ifPresent(errors::addAll);
+            } else {
+                requestValidations.checkExists(operation, FIELD_VALUE).ifPresent(errors::addAll);
+            }
             if (errors.isEmpty()) {
                 requestValidations.checkMaxLength(operation, SERVICE_NAME_MAX_LENGTH, FIELD_VALUE).ifPresent(errors::addAll);
             }
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
@@ -66,7 +66,7 @@ public class ServiceUpdateOperationValidator {
 
     private List<String> validateOpAndPathExistAndNotEmpty(JsonNode operation) {
         List<String> errors = new ArrayList<>();
-        requestValidations.checkIfExistsOrEmpty(operation, FIELD_OP, FIELD_PATH).ifPresent(errors::addAll);
+        requestValidations.checkExistsAndNotEmpty(operation, FIELD_OP, FIELD_PATH).ifPresent(errors::addAll);
         return errors;
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -77,7 +77,11 @@ public ServiceUpdateOperationValidator(RequestValidations requestValidations) {
         if (FIELD_CUSTOM_BRANDING.equals(path)) {
             errors.addAll(checkIfValidJson(operation.get(FIELD_VALUE), FIELD_CUSTOM_BRANDING));
         } else if (FIELD_NAME.equals(path) || path.startsWith(FIELD_SERVICE_NAME_PREFIX)) {
-            requestValidations.checkIfExistsOrEmpty(operation, FIELD_VALUE).ifPresent(errors::addAll);
+            if (FIELD_NAME.equals(path) || path.endsWith('/' + SupportedLanguage.ENGLISH.toString())) {
+                requestValidations.checkIfExistsOrEmpty(operation, FIELD_VALUE).ifPresent(errors::addAll);
```
oswaldquek: I find the method name `checkIfExistsOrEmpty` sorely confusing. It seems like there is no validation error if the field value exists, but there is a validation error the field value is empty. But the method name is suggesting the presence or absence of a validation error for both scenarios. `failValidationIfEmpty` or `validateFieldValueIsNotEmpty` might be better.

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -77,7 +77,11 @@ public ServiceUpdateOperationValidator(RequestValidations requestValidations) {
...
+            } else {
+                requestValidations.checkIfExists(operation, FIELD_VALUE).ifPresent(errors::addAll);
```
oswaldquek: Same comment as above. `checkIfExists` - its not clear whether there should be a validation error if something exists or not.

src/test/java/uk/gov/pay/adminusers/resources/RequestValidationsTest.java
```diff
@@ -41,19 +41,51 @@ public void checkIfExists_shouldFail_whenFieldsAreProvidedWithEmptyValues() thro
...
+        Optional<List<String>> errors = requestValidations.checkIfExists(payload, FIELD_1, FIELD_2);
+
+        assertThat(errors.isPresent(), is(false));
+    }
+
+    @Test
+    public void checkIfExists_shouldSucceed_whenFieldsAreProvidedWithEmptyValues() {
+        ObjectNode payload = JsonNodeFactory.instance.objectNode();
+        payload.put(FIELD_1, "");
+        payload.put(FIELD_2, "");
```
oswaldquek: how about making FIELD_2 have some whitespace, e.g. `" "`?

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceUpdateTest.java
```diff
@@ -232,6 +234,33 @@ public void shouldUpdateName_ServiceNameEn_AndServiceNameCy_whenReplaceServiceNa
...
+                .method("PATCH", Entity.json(jsonPayload));
+
+        assertThat(response.getStatus(), is(200));
+
+        String body = response.readEntity(String.class);
+        JsonPath json = JsonPath.from(body);
+
+        assertThat(json.get("name"), is("old-en-name"));
+        assertEnServiceNameJson("old-en-name", json);
+        assertThat(json.getMap("service_name"), not(hasKey(SupportedLanguage.WELSH.toString())));
```
oswaldquek: ewww... hamcrest... but ok.

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
alexbishop1: You’re right, it’s confusing; it was like that already. We’re going to (in a separate commit) rename this method to `checkExistsAndNotEmpty`, which should be clearer but still be consistent with the other methods.

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
alexbishop1: We will change this to be `checkExists`.

src/test/java/uk/gov/pay/adminusers/resources/RequestValidationsTest.java
alexbishop1: Good idea: we’ll do this.

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceUpdateTest.java
alexbishop1: 🤷‍♂️

PR #299, opened 2018-12-03
BAU: Update pay-java-commons to 1.0.0-f6097986677849386f283bb84072198b57931b1b

rhowe-gds:
Merge the 3 dependabot PRs into one

https://github.com/alphagov/pay-adminusers/pull/298
https://github.com/alphagov/pay-adminusers/pull/296
https://github.com/alphagov/pay-adminusers/pull/289

The docker test harness for spinning up postgres is now in `pay-java-commons` `testing` lib
The Xray initialiser now takes an extra `Optional` parameter for overriding the sampling rules, but we don't need to do any overriding.


pom.xml
```diff
@@ -269,7 +269,19 @@
         <dependency>
             <groupId>uk.gov.pay</groupId>
             <artifactId>testing</artifactId>
-            <version>1.0.0-7</version>
+            <version>1.0.0-95d787ff5931ee0d5e54afd7fbc3834cf772dad4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.json</groupId>
```
oswaldquek: presumably this dependency is unavoidable?

pom.xml
```diff
@@ -269,7 +269,19 @@
...
+            <artifactId>javax.json-api</artifactId>
+            <version>1.1.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish</groupId>
```
oswaldquek: as is this one?

pom.xml
rhowe-gds: Probably not if we do this instead: https://github.com/alphagov/pay-java-commons/pull/36

pom.xml
```diff
@@ -269,7 +269,19 @@
...
-            <version>1.0.0-7</version>
+            <version>1.0.0-95d787ff5931ee0d5e54afd7fbc3834cf772dad4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
+            <version>1.1.4</version>
+            <scope>test</scope>
+        </dependency>
```
alexbishop1: Not certain but I think this might already be covered by the GlassFish dependency

pom.xml
```diff
@@ -269,7 +269,19 @@
...
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.1.4</version>
```
alexbishop1: This being required because `pay-java-commons` includes EclipseLink JPA 2.7.3 rather than full EclipseLink 2.6.4 and upgrading to EclipseLink 2.7.3 means we need this dependency now for Pact tests (see, for example, https://github.com/alphagov/pay-connector/pull/723 for the equivalent change in connector)

pom.xml
```diff
@@ -198,7 +198,7 @@
         <dependency>
             <groupId>uk.gov.pay</groupId>
             <artifactId>utils</artifactId>
-            <version>1.0.0-6</version>
+            <version>1.0.0-f6097986677849386f283bb84072198b57931b1b</version>
```
oswaldquek: this can be replaced with `${pay-java-commons.version}`

pom.xml
```diff
@@ -271,7 +271,7 @@
         <dependency>
             <groupId>uk.gov.pay</groupId>
             <artifactId>testing</artifactId>
-            <version>1.0.0-7</version>
+            <version>1.0.0-f6097986677849386f283bb84072198b57931b1b</version>
```
oswaldquek: same as above

pom.xml
rhowe-gds: So it can. Done.

pom.xml
rhowe-gds: Quite so

PR #310, opened 2018-12-12
PP-4520 Updating a service's name should also update English language service name as well

SandorArpa:
## WHAT
- refactor ServiceUpdater to automagically update English language service name when updating the 'name' field
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
@@ -79,7 +79,8 @@ public class ServiceUpdater {
     }
 
     private BiConsumer<ServiceUpdateRequest, ServiceEntity> updateServiceName() {
-        return (serviceUpdateRequest, serviceEntity) -> serviceEntity.setName(serviceUpdateRequest.valueAsString());
+        return (serviceUpdateRequest, serviceEntity) -> serviceEntity
+                .addOrUpdateServiceName(ServiceNameEntity.from(SupportedLanguage.ENGLISH, serviceUpdateRequest.valueAsString()));
     }
 
     private BiConsumer<ServiceUpdateRequest, ServiceEntity> assignGatewayAccounts() {
--- a/src/test/java/uk/gov/pay/adminusers/fixtures/ServiceDbFixture.java
+++ b/src/test/java/uk/gov/pay/adminusers/fixtures/ServiceDbFixture.java
@@ -51,6 +51,7 @@ public class ServiceDbFixture {
         Service service = Service.from(serviceId, extId, name);
         service.setMerchantDetails(merchantDetails);
         service.setCollectBillingAddress(collectBillingAddress);
+        service.getServiceNames().put("en", service.getName());
         databaseHelper.addService(service, gatewayAccountIds);
 
         return service;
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/pay/adminusers/service/ServiceUpdater.java
```diff
@@ -79,7 +79,11 @@ public Service doUpdateMerchantDetails(String serviceExternalId, UpdateMerchantD
     }
 
     private BiConsumer<ServiceUpdateRequest, ServiceEntity> updateServiceName() {
-        return (serviceUpdateRequest, serviceEntity) -> serviceEntity.setName(serviceUpdateRequest.valueAsString());
+        return (serviceUpdateRequest, serviceEntity) -> {
+            serviceEntity.setName(serviceUpdateRequest.valueAsString());
+            serviceEntity.getServiceNames()
+                    .put(SupportedLanguage.ENGLISH, ServiceNameEntity.from(SupportedLanguage.ENGLISH, serviceUpdateRequest.valueAsString()));
```
alexbishop1: I think for the ORM magic to work, this should update the existing `ServiceNameEntity` if there already is one rather than creating a new one every time, in a similar fashion to `ServiceNameEntity.addOrUpdateServiceName(ServiceNameEntity)`. In fact, we can probably just call that method (which will also update the `name` because of the logic already in there to update the `name` when the `service_name/en` is updated).

src/test/java/uk/gov/pay/adminusers/fixtures/ServiceDbFixture.java
```diff
@@ -51,6 +51,7 @@ public Service insertService() {
         Service service = Service.from(serviceId, extId, name);
         service.setMerchantDetails(merchantDetails);
         service.setCollectBillingAddress(collectBillingAddress);
+        service.getServiceNames().put("EN", service.getName());
```
alexbishop1: This should be lower-case: `"en"`

src/test/java/uk/gov/pay/adminusers/service/ServiceUpdaterTest.java
```diff
@@ -51,6 +52,7 @@ public void before() throws Exception {
     public void shouldUpdateNameSuccessfully() {
         String serviceId = randomUuid();
         String nameToUpdate = "new-name";
+        Map<SupportedLanguage, ServiceNameEntity> languageMap = mock(HashMap.class);
```
alexbishop1: I wouldn’t bother mocking the map: just use a real one and verify its actual contents

src/test/java/uk/gov/pay/adminusers/service/ServiceUpdaterTest.java
```diff
@@ -59,11 +61,14 @@ public void shouldUpdateNameSuccessfully() {
 
         when(serviceDao.findByExternalId(serviceId)).thenReturn(Optional.of(serviceEntity));
         when(serviceEntity.toService()).thenReturn(Service.from());
+        when(serviceEntity.getServiceNames()).thenReturn(languageMap);
 
         Optional<Service> maybeService = updater.doUpdate(serviceId, request);
 
         assertThat(maybeService.isPresent(), is(true));
         verify(serviceEntity, times(1)).setName(nameToUpdate);
+        verify(languageMap, times(1))
```
alexbishop1: Don’t need `times(1)`: it’s the default

PR #313, opened 2018-12-13
PP-4450 add column current_go_live_stage to services

SandorArpa:
## WHAT

- add column current_go_live_stage to services
- set value to 'NOT_STARTED'

with @cobainc0

### DIFF
--- a/src/main/resources/migrations/00046_alter_table_services_add_column_current_go_live_stage.sql
+++ b/src/main/resources/migrations/00046_alter_table_services_add_column_current_go_live_stage.sql
@@ -1,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_services_add_column_current_go_live_stage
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00046_alter_table_services_add_column_current_go_live_stage.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter_table_services_add_column_current_go_live_stage
+ALTER TABLE services ADD COLUMN current_go_live_stage VARCHAR(50) NULL;
+UPDATE services SET current_go_live_stage = 'NOT_STARTED' WHERE current_go_live_stage IS NULL;
+
```
alexbishop1: Probably want to add a `DEFAULT` (in a third statement) so that services created from now on will get an appropriate value before code changes to do this go in (or we’ll have to do another round of backfilling)

PR #317, opened 2018-12-17
PP-4450 Add validation for current_go_live_stage update operation

SandorArpa:
## WHAT

- add operation validation
- add unit tests for validator

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
@@ -2,6 +2,7 @@ package uk.gov.pay.adminusers.resources;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableMap;
+import uk.gov.pay.adminusers.model.GoLiveStage;
 import uk.gov.pay.adminusers.validations.RequestValidations;
 import uk.gov.pay.commons.model.SupportedLanguage;
 
@@ -9,6 +10,7 @@ import javax.inject.Inject;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -312,4 +312,83 @@ public void shouldFail_updatingServiceName_whenValueIsBoolean() {
...
+    @Test
+    public void shouldFail_updatingCurrentGoLiveStage_whenValueIsEmptyString() {
+        ObjectNode payload = mapper.createObjectNode();
+        payload.put("path", "current_go_live_stage");
+        payload.put("op", "replace");
+        payload.put("value", "");
+        List<String> errors = serviceUpdateOperationValidator.validate(payload);
+        assertThat(errors.size(), is(1));
+        assertThat(errors, hasItem("Field [value] is required"));
+    }
```
alexbishop1: Nice set of tests

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -102,6 +108,17 @@ public ServiceUpdateOperationValidator(RequestValidations requestValidations) {
             if (errors.isEmpty()) {
                 requestValidations.checkIsStrictBoolean(operation, FIELD_VALUE).ifPresent(errors::addAll);
             }
+        } else if (FIELD_CURRENT_GO_LIVE_STAGE.equals(path)) {
+            requestValidations.checkExistsAndNotEmpty(operation, FIELD_VALUE).ifPresent(errors::addAll);
+            if (errors.isEmpty()) {
+                requestValidations.checkIsString(
+                        format("Field [%s] must be one of %s", FIELD_VALUE, goLiveStageEnumSet),
```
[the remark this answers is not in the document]
SandorArpa: It's because we have a custom error message and not that of the original checkIsString

src/main/java/uk/gov/pay/adminusers/validations/RequestValidations.java
```diff
@@ -58,6 +64,14 @@
         }
         return errors.isEmpty() ? Optional.empty() : Optional.of(errors);
     }
+    
+    public Optional<List<String>> isValidEnumValue(JsonNode payload, Class enumClass, String field) {
+        String value = payload.get(field).asText();
+        if (EnumSet.allOf(enumClass).stream().noneMatch(constant -> constant.toString().equals(value))) {
```
alexbishop1: We could have this method take an `EnumSet` instead of the enum `Class` object and avoid having to recreate it i.e. make the method signature `isValidEnumValue(JsonNode payload, EnumSet<?> enumClass, String field)`

src/main/java/uk/gov/pay/adminusers/validations/RequestValidations.java
```diff
@@ -58,6 +64,14 @@
         }
         return errors.isEmpty() ? Optional.empty() : Optional.of(errors);
     }
+    
+    public Optional<List<String>> isValidEnumValue(JsonNode payload, EnumSet enumSet, String field) {
```
alexbishop1: Should really be `EnumSet<?>`

src/main/java/uk/gov/pay/adminusers/validations/RequestValidations.java
```diff
@@ -58,6 +64,14 @@
...
+        String value = payload.get(field).asText();
+        if (enumSet.stream().noneMatch(constant -> constant.toString().equals(value))) {
```
alexbishop1: It would be safer to use `.name()` here because it will always give the exact name, whereas `toString()` can be overridden (you’ll have to add the generic type parameter detailed in my other comment for this to work)

PR #319, opened 2018-12-18
BAU: Reduce the number of warnings from IntelliJ

rhowe-gds:
Hopefully a no-op change. Address a bunch of IntelliJ's gripes (see individual commits)


rhowe-gds:
This has gotten too stale. Will redo in chunks.

PR #321, opened 2018-12-20
PP-4450 Add property to service model

cobainc0:
- adds resource test for `current_go_live_stage` get and update endpoints

Tests can be found:
`ServiceResourceUpdateGoLiveStageTest`
`ServiceResourceFindTest`

with: @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Service.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Service.java
@@ -176,6 +176,7 @@ public class Service {
         this.collectBillingAddress = collectBillingAddress;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/adminusers/model/Service.java
```diff
@@ -176,6 +176,7 @@ public void setCollectBillingAddress(boolean collectBillingAddress) {
         this.collectBillingAddress = collectBillingAddress;
     }
 
+    @JsonProperty("current_go_live_stage")
```
DanailMinchev: Well spotted 👍

PR #324, opened 2018-12-24
PP-4449 Add role permissions for Go Live stage

cobainc0:
- new migration adds role permissions for admins only

with: @stephencdaly

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/resources/UserResourceAuthenticationTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/resources/UserResourceAuthenticationTest.java
@@ -51,7 +51,7 @@ public class UserResourceAuthenticationTest extends IntegrationTest {
                 .body("disabled", is(false))
                 .body("_links", hasSize(1))
                 .body("service_roles[0].role.name", is("admin"))
-                .body("service_roles[0].role.permissions", hasSize(35)); //we could consider removing this assertion if the permissions constantly changing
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/adminusers/resources/UserResourceAuthenticationTest.java
```diff
@@ -51,7 +51,7 @@ public void shouldAuthenticateUser_onAValidUsernamePasswordCombination() throws
                 .body("disabled", is(false))
                 .body("_links", hasSize(1))
                 .body("service_roles[0].role.name", is("admin"))
-                .body("service_roles[0].role.permissions", hasSize(35)); //we could consider removing this assertion if the permissions constantly changing
+                .body("service_roles[0].role.permissions", hasSize(37));
```
DanailMinchev: As discussed we were wondering why it is not `38` as the total amount of permissions in the database.
We have `users-global:create` permission here https://github.com/alphagov/pay-adminusers/blob/5aeaa9c5382585a01e5d6d3373b0b3bdf53e795c/src/main/resources/config/initdata/permissions.csv#L4
Which is actually not linked/added to `admin` role here https://github.com/alphagov/pay-adminusers/blob/5aeaa9c5382585a01e5d6d3373b0b3bdf53e795c/src/main/resources/config/initdata/role-permissions.csv#L24
and the tests are executed against `admin` role, that is why we have `-1` permission from the total ones.

PR #335, opened 2019-01-03
PP-4539 Create components to persist stripe agreement information

danworth:
- Add `StripeAgreementDao` to manage persistence of stripe agreements.
- Add `StripeAgreementEntity`.
- Add tests for `StripeAgreementDao` including necessary fixtures and methods to
`DatabaseTestHelper` class.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/StripeAgreementEntity.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/StripeAgreementEntity.java
@@ -1,0 +1,56 @@
+package uk.gov.pay.adminusers.persistence.entity;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/persistence/entity/StripeAgreementEntity.java
```diff
@@ -0,0 +1,56 @@
...
+    private int id;
+
+    @Column(name = "service_id")
+    private int serviceId;
+
+    @Column(name = "ip_address")
+    private String ipAddress;
+
+    @Column(name = "agreement_time", columnDefinition = "TIMESTAMP WITHOUT TIME ZONE NOT NULL")
+    private LocalDateTime agreementTime;
```
DanailMinchev: I guess when we set the value for this property we are going to use `UTC`, for example:
```
LocalDateTime.now(ZoneOffset.UTC)
```

PR #341, opened 2019-01-04
PP-4449 Add Pact States for Request To Go Live users

DanailMinchev:
## WHAT

Add needed Pact states for Request To Go Live users used in https://github.com/alphagov/pay-selfservice/pull/1072

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/pact/ProviderContractTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/pact/ProviderContractTest.java
@@ -140,6 +140,66 @@ public class ProviderContractTest {
         createUserWithinAService("7d19aff33f8948deb97ed16b2912dcd3", "existing-user", "password");
     }
 
+    @State("a user exists with the given external id rtglNotStartedUserExtId")
+    public void aUserExistsWithGivenExternalIdAndNotStartedGoLiveStage() {
+        createUserWithinAService("rtglNotStartedUserExtId", "rtglNotStarted", "password");
+    }
+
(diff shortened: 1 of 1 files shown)


DanailMinchev:
Initially I was thinking to stick with lowercase variant (to be consistent with external ids format) and it is already merged on https://github.com/alphagov/pay-adminusers/blob/a1e14b6d7e8cbd5e70de6f04f3ac861ad40ed834/src/test/java/uk/gov/pay/adminusers/pact/ProviderContractTest.java#L151
We can try to change it to external ids with camelCase to see if it is not a problem, I can create another PR to change the existing ids to camel case starting here https://github.com/alphagov/pay-adminusers/blob/a1e14b6d7e8cbd5e70de6f04f3ac861ad40ed834/src/test/java/uk/gov/pay/adminusers/pact/ProviderContractTest.java#L151 and then change this PR or we can address that as part of Cypress refactoring (as discussed with @stephencdaly probably we want to minimise the usage of https://github.com/alphagov/pay-selfservice/blob/d221e6a8fe9b229c048849c0f6b37946ae753588/test/fixtures/config/self_service_user.json and static fixtures which are executed against Pact)
Let me know what you think is better, thanks

danworth:
If its a simple change to make them camel case then I think the improved readability is worth it. If it requires greater refactoring then I'm happy to approve as-is if we can capture it when the refactoring of Cypress tests happens. I'll let you decide.

DanailMinchev:
I think it's better to change them now and see if it will break something, should not be a problem, I will create this as part of this PR actually as the ids are not used in selfservice yet

PR #344, opened 2019-01-09
PP-4594 Add last logged in at property

DanailMinchev:
## WHAT

- Added `lastLoggedInAt` property to `UserEntity`
- Added `lastLoggedInAt` property to `User`

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/UserServices.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/UserServices.java
@@ -172,8 +172,9 @@ public class UserServices {
     }
 
     @Transactional
-    public Optional<User> authenticateSecondFactor(String externalId, int code, ZonedDateTime now) {
+    public Optional<User> authenticateSecondFactor(String externalId, int code) {
         logger.debug("OTP attempt - user_id={}", externalId);
+        ZonedDateTime now = ZonedDateTime.now(ZoneId.of("UTC"));
         return userDao.findByExternalId(externalId)
                 .map(userEntity -> {
                     if (userEntity.isDisabled()) {
--- a/src/main/java/uk/gov/pay/adminusers/service/UserServices.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/UserServices.java
@@ -109,7 +109,6 @@ public class UserServices {
     }
 
     /**
-     *
      * @param externalIds
      * @return A {@link List} of {@link User} or an empty {@link List} otherwise
      */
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/adminusers/service/UserServices.java
```diff
@@ -173,7 +172,7 @@ public UserServices(UserDao userDao,
     }
 
     @Transactional
-    public Optional<User> authenticateSecondFactor(String externalId, int code) {
+    public Optional<User> authenticateSecondFactor(String externalId, int code, ZonedDateTime now) {
```
SandorArpa: Why are we passing in the value instead of initialising it locally?

DanailMinchev: Initially I initialised it locally, but later realised that to be able to test it, we need to mock `ZonedDateTime.now` for example here https://github.com/alphagov/pay-adminusers/blob/e9f7ce496bf24d14860206be439b5a50213a4a3d/src/test/java/uk/gov/pay/adminusers/service/UserServicesTest.java#L437 and then assert on a mocked value.
We cannot mock static methods without introducing something like https://github.com/powermock/powermock (at least I was unable to find other solution with quick research).

That is why I have decided to move the value as part of the method call to be able to test it properly.

Please let me know if you have any other ideas how to test that, thanks

SandorArpa: You could test it with a relative value, eg:

`ZonedDateTime now = ZonedDateTime.now().minusSeconds(10);
...
 assertThat(user.getLastLoggedInAt().isAfter(now), is(true));`

DanailMinchev: Good point! I will try that

DanailMinchev: As discussed with @alexbishop1 and @SandorArpa we currently use same approach as @SandorArpa suggested using https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/matcher/ZoneDateTimeAsStringWithinMatcher.java#L24

Also, we discussed that ideally we should inject `Clock` class into the service and then manipulate it for testing as suggested by @alexbishop1 but we agreed that it is not the current pattern which we use, so we will not do that for now

src/main/java/uk/gov/pay/adminusers/service/UserServices.java
```diff
@@ -175,6 +174,7 @@ public UserServices(UserDao userDao,
     @Transactional
     public Optional<User> authenticateSecondFactor(String externalId, int code) {
         logger.debug("OTP attempt - user_id={}", externalId);
+        ZonedDateTime now = ZonedDateTime.now(ZoneId.of("UTC"));
```
SandorArpa: You could also use `ZonedDateTime.now(ZoneOffset.UTC)`

DanailMinchev: Yes, looks like we use `ZonedDateTime.now(ZoneId.of("UTC"))` a bit more in the project

PR #346, opened 2019-01-10
PP-4607 Remove deprecated fields in User response

stephencdaly:
- 'services', 'role', 'permissions' and 'gateway_account_ids' are all deprecated as this information is now under 'service_roles'
- selfservice no longer uses these fields so they can be removed


src/test/java/uk/gov/pay/adminusers/service/ServiceRoleUpdaterTest.java
```diff
@@ -129,7 +130,9 @@ public void shouldReturnUpdatedUser_whenUpdatingServiceRoleSuccess() {
 
         Optional<User> userOptional = serviceRoleUpdater.doUpdate(EXISTING_USER_EXTERNAL_ID, serviceExternalId, role);
         assertTrue(userOptional.isPresent());
-        assertThat(userOptional.get().getRole().getId(), is(10));
+        // TODO: this looks like a bug in the updater, should only be 1 service role
```
stephencdaly: I've looked into this and it is an error in what the updater returns rather than what's updated. I'll fix it in a separate PR.

PR #350, opened 2019-01-14
BAU: More IntelliJ-inspired code cleanups

rhowe-gds:
* Remove redundant declarations
* Construct lists more efficiently
* Neater functional constructs
* Better use of Optionals
* Use try-with-resources for managing JDBC Connection objects
* More efficient String handling


rhowe-gds:
That was just how IntelliJ chose to do it, and I blindly followed

PR #352, opened 2019-01-14
PP-4581 Add table govuk_pay_agreements

SandorArpa:
- add script to create table
(new table (govuk_pay_agreements) (columns: service_id (integer -> FK to services.id) NOT NULL, email (varchar 254) NOT NULL, agreement_time (timestamp without time zone) NOT NULL)

### DIFF
--- a/src/main/resources/migrations/00050_create_table_govuk_pay_agreements.sql
+++ b/src/main/resources/migrations/00050_create_table_govuk_pay_agreements.sql
@@ -1,0 +1,10 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_govuk_pay_agreements
+CREATE TABLE govuk_pay_agreements (
+  id SERIAL PRIMARY KEY,
+  service_id INT NOT NULL UNIQUE REFERENCES services (id),
+  email VARCHAR(254) NOT NULL,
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00050_create_table_govuk_pay_agreements.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_govuk_pay_agreements
+CREATE TABLE govuk_pay_agreementss (
```
alexbishop1: Only one “s”!

src/main/resources/migrations/00050_create_table_govuk_pay_agreements.sql
```diff
@@ -0,0 +1,10 @@
...
+  id SERIAL PRIMARY KEY,
+  service_id INT NOT NULL UNIQUE REFERENCES services (id),
+  email VARCHAR(254) NOT NULL,
+  agreement_time TIMESTAMP WITHOUT TIME ZONE NOT NULL
```
alexbishop1: I think we want `TIMESTAMP WITH TIME ZONE` (the time zone being UTC)

PR #353, opened 2019-01-14
BAU: Logging fixes - incorrect formatting placeholders

rhowe-gds:
lgtm.com spotted a couple of cases where our log messages' placeholders were incorrect. Fix them up so that our logs are what we expect.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/app/healthchecks/MigrateToInitialDbState.java
+++ b/src/main/java/uk/gov/pay/adminusers/app/healthchecks/MigrateToInitialDbState.java
@@ -22,7 +22,7 @@ public class MigrateToInitialDbState extends ConfiguredCommand<AdminUsersConfig>
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/adminusers/app/healthchecks/MigrateToInitialDbState.java
```diff
@@ -38,7 +38,7 @@ protected void run(Bootstrap<AdminUsersConfig> bootstrap, Namespace namespace, A
             resultSet.close();
             statement.close();
         } catch (SQLException e) {
-            logger.error("Error during initial DB setup", e.getMessage());
+            logger.error("Error during initial DB setup", e);
```
alexbishop1: Putting the whole exception tends to toss each line of the stack trace into Sumo Logic as a separate log message, so it’s possible `e.getMessage()` was considered. But, meh, change it, it’s not like we expect this error to ever actually happen and a stack trace would probably be useful if it did.

rhowe-gds: gah, had to rebase

PR #356, opened 2019-01-15
PP-4447 Refactoring of stripe agreement request.

danworth:
Some refactoring to adopt standards encouraged by the Dropwizard documentation.
- Change `ServiceResource.createStripeAgreement()` to accept a `@NotNull` and
`@Valid` `StripeAgreementRequest`. This allows Jersey and Hibernate to handle
the validation in a resuable and easy to read manner, and declutters the logic
within the Resource method.
- Throw appropriate `WebApplicationExceptions` rather than creating a `Response`
and setting the status code to that particular error.
- Changed the return type of `ServiceResource.getStripeAgreement` to
`StripeAgreement`. Recommended by DropWizard documentation to return domain types
and avoid using `Response` where possible.
- Create `StripeAgreementRequest` to permit validation of request
by Jersey and Hibernate rather than relying upon validating a String within
the method body. The `ipAddress` is held as a String rather than a more complex
type representing a URI/ipAddress for simplicity, since we store and retrieve it
as a String without any manipulation.
- Create custom hibernate validation annotation for Ip address. This can be
reused across the project or added to pay-commons. Considers both Ipv4 and Ipv6.
- Delete `StripeAgreementRequestValidator` since this is superseded by the
new `@ValidIpAddress` validation. Remove reference and injection of this class
in various places.

**WHAT**
The commit message hopefully explains most of it. I learnt a lot during firebreak and so I've refactored a story I implemented just before with the aim of using Dropwizard recommendations. IMO it is easier to understand and I think will make changes easier to implement. Some points for consideration:
- Hibernate validations will return an error that references the Java attribute name rather than the serialised json name, e.g. `ipAddress` rather than `ip_address`. Might be possible to alter this but not obvious since Jackson which understands the serialisation of a custom Type isn't involved in returning the Exception (as far as I can see). Since the message is intended to be read by a human I don't see this as a blocker, opinions welcomed :)
- Rather than have the `StripeAgreementService` handle the check for the existing agreement or whether the service exists (it cannot over write it if it does exist), I feel this would make it harder to see the logic that the Resource is implementing and the various results a client could receive. I'm also keen on the idea that setters should return `void` rather than asking a setter to try something and let you know, hence the Resource first uses the service to confirm it isn't set before telling it to create it (returning a sensible error to the client if it is set). This keeps some logic in the Resource method but I feel the Resources should be used to coordinate activity between services and handle the response side of things. The logic that remains should be lean and clearly show what that end-point is trying to achieve. Also, that services shouldn't really have to know about each other because I've seen this creating circular dependencies which cause problems when trying to inject them). Happy to hear different opinions and open to different views, I'm by no means an expert, just trying to stitch together things I've learnt from various sources.

This is a bit of an example to gather opinions on a real life example. Happy to hear any comments.


src/main/java/uk/gov/pay/adminusers/model/StripeAgreementRequest.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.adminusers.resources.ValidIpAddress;
+
+public class StripeAgreementRequest {
+    
+    @ValidIpAddress
+    private final String ipAddress;
+    
+    @JsonCreator
+    public StripeAgreementRequest(@JsonProperty("ip_address") String ipAddress) {
```
alexbishop1: Can we use `JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)` instead and not have to explicitly snake-case each property?

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -227,44 +227,32 @@ public Response removeUserFromService(@PathParam("serviceExternalId") String ser
...
+        if (stripeAgreementService.findStripeAgreementByServiceId(service.getId()).isPresent()) {
+            throw new WebApplicationException("Stripe agreement information is already stored for this service",
+                    Status.CONFLICT);
         }
 
-        stripeAgreementService.doCreate(
-                maybeService.get().getId(),
-                payload.get(FIELD_IP_ADDRESS).asText(),
-                LocalDateTime.now(ZoneOffset.UTC));
+        stripeAgreementService.doCreate(service.getId(), stripeAgreementRequest, LocalDateTime.now(ZoneOffset.UTC));
```
alexbishop1: Completely unrelated to the purpose of the PR but I’m surprised we’re using a `LocalDateTime`. We want an absolute time (for which we usually use `ZonedDateTime` though I sometimes think we should probably use `Instant` instead).

src/main/java/uk/gov/pay/adminusers/service/StripeAgreementService.java
```diff
@@ -28,8 +29,8 @@ public StripeAgreementService(StripeAgreementDao stripeAgreementDao) {
                 .map((StripeAgreementEntity::toStripeAgreement));
     }
     
-    public void doCreate(int serviceId, String ipAddress, LocalDateTime agreementTime) {
+    public void doCreate(int serviceId, StripeAgreementRequest stripeAgreementRequest, LocalDateTime agreementTime) {
```
alexbishop1: I think I’d prefer if we didn’t pass the `StripeAgreementRequest` in here and instead passed in just the IP address (which could be stored as a `InetAddress` or something rather than a `String`). Ideally, a service shouldn’t really know that it’s part of a web application.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceStripeAgreementTest.java
```diff
@@ -76,9 +75,23 @@ public void shouldReturn_400_whenProvidedInvalidIPAddress() {
                 .body(payload)
                 .post(format("/v1/api/services/%s/stripe-agreement", service.getExternalId()))
                 .then()
-                .statusCode(400)
+                .statusCode(422)
```
stephencdaly: where does the 422 status code come from for this?

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -227,30 +229,21 @@ public Response removeUserFromService(@PathParam("serviceExternalId") String ser
...
+            throw new WebApplicationException("Stripe agreement information is already stored for this service",
+                    Status.CONFLICT);
         }
 
-        stripeAgreementService.doCreate(
-                maybeService.get().getId(),
-                payload.get(FIELD_IP_ADDRESS).asText(),
+        stripeAgreementService.doCreate(service.getId(),
+                InetAddress.getByName(stripeAgreementRequest.getIpAddress()),
                 LocalDateTime.now(ZoneOffset.UTC));
```
stephencdaly: Was the question about LocalDateTime resolved?

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceStripeAgreementTest.java
danworth: The Hibernate validation which is invoked with the `@ValidIpAddress` returns 422 if a parameter fails validation because it is an invalid entity. This seems appropriate given its description:
https://httpstatuses.com/422

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
danworth: No, I think that will need changing, but rather than confuse commits and PRs I thought that could be done afterwards to keep this refactoring clean.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceStripeAgreementTest.java
stephencdaly: Makes sense. In that case, I think you just need to rename this test method and the test below it.

PR #357, opened 2019-01-15
PP-4581 Implement GovUkPayAgreementDao

SandorArpa:
- implement GovUkPayAgreementDao
- add relevant tests

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/persistence/dao/GovUkPayAgreementDaoTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/persistence/dao/GovUkPayAgreementDaoTest.java
@@ -1,0 +1,86 @@
+package uk.gov.pay.adminusers.persistence.dao;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/adminusers/persistence/dao/GovUkPayAgreementDaoTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+        GovUkPayAgreementEntity anotherEntity = new GovUkPayAgreementEntity(serviceId,
+                "someone.else@example.org",
+                LocalDateTime.now());
+        
+        try {
+            agreementDao.persist(anotherEntity);
+            fail();
+        } catch (RollbackException e) {
+            assertTrue(e.getMessage().contains("Key (service_id)=(" + serviceId + ") already exists."));
```
stephencdaly: Would it be better to use the ExpectedException rule to assert that the exception is thrown in this case?
If the exception wasn't thrown it wouldn't reach this line so the test would still pass.

SandorArpa: We can

PR #363, opened 2019-01-17
PP-4615 Use ZonedDateTime for stripe agreement timestamp

stephencdaly:
- Were previously using LocalDateTime, but should use ZonedDateTime for
  consistency


stephencdaly:
This was committed with the wrong JIRA ID, it should have been PP-4539

PR #365, opened 2019-01-17
PP-4539 Make StripeAgreementDao search by service external id

stephencdaly:
- Previously was searching by service ID which required first looking
  up the service in a separate query
- Refactored code to move more logic into the service from the resource


stephencdaly:
Me and @danworth have had a bit of a chat about responsibilities of the service and resource layers and the best way to propagate and map exceptions to error responses, which we're going to talk about more on Monday.
I'm going to merge this pull request but probably put in a new one with whatever changes to the resource and service we agree on.

danworth:
👍 Really enjoyed our chat. I don't think its easy to decide what is best and each options provides various benefits and concessions. Look forward to picking up next week.

PR #367, opened 2019-01-18
PP-4660, PP-4616: Upgrade pay-java-commons

rhowe-gds:
We can use an amended logformat string instead of `PayLoggerFactory` and its proxied `Logger`.

By removing this, we are able to use the shared implementation of `LoggerFactory` in `pay-java-commons`' `utils`.

Also this update brings in a simpler way of creating `ApplicationStartupDependentResourceChecker` objects, so we can drop our lambda (the class will supply its own).


rhowe-gds:
Superseded by https://github.com/alphagov/pay-adminusers/pull/372

PR #371, opened 2019-01-21
PP-4607 Update API documentation for user response

stephencdaly:
- Remove fields that no longer exist
- Update documentation to include all user fields that do exists
- Remove duplication in docs and reference a single place for the user
  object specification

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -1,7 +1,85 @@
-#API Specification
+# API Specification
 These are the endpoints and methods available for managing users on GOV.UK Pay.
 
-## POST /v1/api/users
+### The user object
+
+| Field                            | always present | Description                                                        |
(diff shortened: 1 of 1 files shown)


docs/api_specification.md
```diff
@@ -35,39 +113,13 @@ Content-Type: application/json
 ### Response example
 
 ```
-201 OK
+200 OK
```
DanailMinchev: Is it not `201 OK` https://github.com/alphagov/pay-adminusers/blob/4cfd993b5652300f45afd0ef40b33c574a091c3b/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java#L126 ?

stephencdaly: Good spot, I must have been looking at the wrong one, I'll fix

PR #373, opened 2019-01-21
PP-4451 Organisation name validation

cobainc0:
- Added organisation name validation for maximum characters
- Fixed `ServiceRequestValidatorTest` tests to properly assert on exception

with @DanailMinchev

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
@@ -20,6 +20,8 @@ public class ServiceRequestValidator {
     static final String FIELD_MERCHANT_DETAILS_ADDRESS_POSTCODE = "address_postcode";
     static final String FIELD_MERCHANT_DETAILS_ADDRESS_COUNTRY = "address_country";
     static final String FIELD_MERCHANT_DETAILS_EMAIL = "email";
+
+    private static final int FIELD_MERCHANT_DETAILS_NAME_MAX_LENGTH = 255;
     private static final int FIELD_MERCHANT_DETAILS_EMAIL_MAX_LENGTH = 255;
 
     private final RequestValidations requestValidations;
@@ -34,7 +36,7 @@ public class ServiceRequestValidator {
     Optional<Errors> validateUpdateAttributeRequest(JsonNode payload) {
         List<String> errors = new ArrayList<>();
         if (!payload.isArray()) {
-            errors  = serviceUpdateOperationValidator.validate(payload);
+            errors = serviceUpdateOperationValidator.validate(payload);
         } else {
             for (JsonNode updateOperation : payload) {
                 errors.addAll(serviceUpdateOperationValidator.validate(updateOperation));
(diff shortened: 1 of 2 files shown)


DanailMinchev:
@stephencdaly yes, I totally agree, I actually like this approach more than the other one where we validate the `JsonNode` manually.

I was using it before in my last place, for example https://github.com/DanailMinchev/spring-boot-auth/blob/master/src/main/java/com/example/user/validation/definition/UserDefinitionUniqueUserEmailValidator.java using `Hibernate Validator` with custom annotation.

Basically as you said we have two options:

1. To use `Hibernate Validator` and `@Valid` annotation on resource level

https://github.com/alphagov/pay-adminusers/blob/1d457e4e54501fb4f7d278a223bf34281e9a8dcb/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java#L242

and

https://github.com/alphagov/pay-adminusers/blob/1d457e4e54501fb4f7d278a223bf34281e9a8dcb/src/main/java/uk/gov/pay/adminusers/model/StripeAgreement.java#L9

2. To use dedicated class which will validate `JsonNode` 

https://github.com/alphagov/pay-adminusers/blob/1d457e4e54501fb4f7d278a223bf34281e9a8dcb/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java#L264

and

https://github.com/alphagov/pay-adminusers/blob/1d457e4e54501fb4f7d278a223bf34281e9a8dcb/src/main/java/uk/gov/pay/adminusers/resources/GovUkPayAgreementRequestValidator.java#L11

I think the second option also has multiple ways to implement it depending what is received on resource level and how it is handled:

- we can use `JsonNode` and functions

https://github.com/alphagov/pay-adminusers/blob/1d457e4e54501fb4f7d278a223bf34281e9a8dcb/src/main/java/uk/gov/pay/adminusers/validations/RequestValidations.java#L59

- we can use `Map<String, String>` and base validation class like `ApiValidation` with custom class for specific validations

https://github.com/alphagov/pay-direct-debit-connector/blob/2d9296cdf3a409fcb08686cc6b32cee43dc39916/src/main/java/uk/gov/pay/directdebit/mandate/resources/MandateResource.java#L51

and

https://github.com/alphagov/pay-direct-debit-connector/blob/2d9296cdf3a409fcb08686cc6b32cee43dc39916/src/main/java/uk/gov/pay/directdebit/mandate/api/CreateMandateRequestValidator.java#L10

and

https://github.com/alphagov/pay-direct-debit-connector/blob/2d9296cdf3a409fcb08686cc6b32cee43dc39916/src/main/java/uk/gov/pay/directdebit/common/validation/ApiValidation.java#L18

I think using `Hibernate Validator` is much more consistent way (at least it is standard) and we can share the validations more easily (also to rely on returning error codes and mapping them to various messages in case we want to internationalise our apps at some point).

On the other hand when using raw `JsonNode` and `Map` you can follow the validation logic "line by line".

I think we were discussing this before (we had learning sessions as well), but we were unable to agree on a solution I think as it needs major refactoring and moving to only one way, hence we try to keep it consistent with what was implemented already, hence we have 3 ways to validate the requests, unfortunately :(

PR #377, opened 2019-01-22
PP-4581 add API documentation

SandorArpa:
- add API specs for the new endpoint

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -939,7 +939,7 @@ Content-Type: application/json
 }
 ```
 
-### Request body description
+#### Request body description
 
 | Field                    | required | Description                                                    | Supported Values     |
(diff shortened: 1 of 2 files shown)


docs/api_specification.md
```diff
@@ -974,6 +974,54 @@ Content-Type: application/json
...
+
+```
+200 OK
+```
+
+-----------------------------------------------------------------------------------------------------------
+
+### GET /v1/api/services/`{serviceExternalId}`/govuk-pay-agreement
+
+This endpoint retrieves email address of the user's and timestamp that the GOV.UK Pay terms were accepted on for the service.
```
danworth: Minor point but perhaps slightly better worded as `This endpoint retrieves the email address of the user and timestamp...` or `This endpoint retrieves the user's email address and timestamp...`

PR #378, opened 2019-01-22
PP-4615 Add documentation for send live service email endpoint

stephencdaly:


docs/api_specification.md
```diff
@@ -974,6 +974,22 @@ Content-Type: application/json
 
 -----------------------------------------------------------------------------------------------------------
 
+### POST /v1/api/services/`{serviceExternalId}`/send-live-email
+
+This endpoint will send an email to the user who signed the agreement with GOV.UK Pay for the service informing them that their service is now live.
```
alexbishop1: Maybe reference the endpoint that’s used to record the email address of the user who signed the agreement here?

PR #379, opened 2019-01-23
PP-4539 Fix responses for stripe agreement endpoints

stephencdaly:
- Success response for POST endpoint is now a 201
- Fixes 500 errors thrown when there are issues with the body of the POST request. Now returns 422 errors with appropriate messages.
- Fixes body of conflict responses to match existing pattern.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
+++ b/src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
@@ -1,6 +1,6 @@
 package uk.gov.pay.adminusers.exception;
 
-public class ServiceNotFoundException extends Throwable {
+public class ServiceNotFoundException extends NotFoundException {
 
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/adminusers/exception/ServiceNotFoundException.java
```diff
@@ -1,6 +1,6 @@
 package uk.gov.pay.adminusers.exception;
 
-public class ServiceNotFoundException extends Throwable {
+public class ServiceNotFoundException extends NotFoundException {
```
stephencdaly: Making this match how we do it in direct-debit-connector by having the ServiceNotFoundException extend NotFoundException then just having an ExceptionMapper for the NotFoundException.

alexbishop1: Yeah, `extends Throwable` is just… _yeesh_.

PR #383, opened 2019-01-25
PP-4614 Add validation for update merchant details patch request

stephencdaly:
- Add validation for all possible paths that can be used to update
  service merchant details
- Add tests and extract methods in test class to avoid some duplication

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
@@ -23,6 +23,14 @@ import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_COLLECT_BILLING
 import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_CURRENT_GO_LIVE_STAGE;
 import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_CUSTOM_BRANDING;
 import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_GATEWAY_ACCOUNT_IDS;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_ADDRESS_CITY;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_ADDRESS_COUNRTY;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_ADDRESS_LINE_1;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_ADDRESS_LINE_2;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_ADDRESS_POSTCODE;
+import static uk.gov.pay.adminusers.service.ServiceUpdater.FIELD_MERCHANT_DETAILS_EMAIL;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -149,7 +170,24 @@ public ServiceUpdateOperationValidator(RequestValidations requestValidations) {
         }
         return errors;
     }
-    
+
+    private List<String> validateNotNullStringValueWithMaxLength(JsonNode operation, boolean allowEmpty, int maxLength) {
+        List<String> errors = new ArrayList<>();
+        if (allowEmpty)
```
SandorArpa: we use curly brackets with one liners, too

src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -15,23 +15,11 @@
 
 public class ServiceUpdateOperationValidatorTest {
 
+    private static final String GO_LIVE_STAGE_INVALID_ERROR_MESSAGE = "Field [value] must be one of [NOT_STARTED, ENTERED_ORGANISATION_NAME, CHOSEN_PSP_STRIPE, CHOSEN_PSP_WORLDPAY, CHOSEN_PSP_SMARTPAY, CHOSEN_PSP_EPDQ, TERMS_AGREED_STRIPE, TERMS_AGREED_WORLDPAY, TERMS_AGREED_SMARTPAY, TERMS_AGREED_EPDQ, DENIED, LIVE]";
+    
     private final ObjectMapper mapper = new ObjectMapper();
     private final ServiceUpdateOperationValidator serviceUpdateOperationValidator = new ServiceUpdateOperationValidator(new RequestValidations());
 
-    @Test
```
stephencdaly: These 2 tests I've just moved down the class a bit so that the common tests are grouped together at the top followed by the service name tests. They're now at line 53

src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -134,7 +107,10 @@ public void shouldFail_whenUpdateServiceName_whenServiceNameFieldPresentAndItIsT
 
     @Test
     public void shouldFail_whenUpdateServiceName_whenPathContainsUnsupportedLanguage() {
-        ImmutableMap<String, String> payload = ImmutableMap.of("path", "service_name/xx", "op", "replace", "value", "example-name");
+        ImmutableMap<String, String> payload = ImmutableMap.of(
+                "path", "service_name/xx",
+                "op", "replace",
+                "value", "example-name");
```
stephencdaly: Not extracted any of this logic as it is quite a specific test

src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -152,21 +128,9 @@ public void shouldFail_whenUpdateRedirectToService_whenOpIsNotReplace() {
...
-        ObjectNode payload = mapper.createObjectNode();
-        payload.put("path", "redirect_to_service_immediately_on_terminal_state");
-        payload.put("value", true);
-
-        List<String> errors = serviceUpdateOperationValidator.validate(payload);
-
-        assertThat(errors.size(), is(1));
-        assertThat(errors, hasItem("Field [op] is required"));
-    }
-
```
stephencdaly: Removed this one because we already have a test checking that errors are thrown when the "op" and "value" fields are missing. This is common logic so don't need a test for each path.

PR #386, opened 2019-01-30
PP-4597 Use common JSON serializer for date time

georgeracu:
- Replaced custom JSON date time serializer with the one from pay java commons
- Upgraded pay-java-commons version to use the new serializer


src/test/java/uk/gov/pay/adminusers/utils/DateTimeUtilsTest.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.pay.adminusers.utils;
+
+import org.junit.Test;
+
```
gidsg: very minot but can we remove the whitespace between these imports?

PR #433, opened 2019-03-19
PP-4895: Upgrade Alpine & Java

oswaldquek:
Upgrade to our OpenJDK base image based on Alpine 3.9 with OpenJDK 8u201


sfount:
It looks like this should point to `PP-4895` -> `PP-4985` -- comment for anyone hunting this down later.

PR #434, opened 2019-03-19
PP-4946 Added `groovy-json` library

kbottla:
## WHAT YOU DID

Added `groovy-json` library - Prep for Java 11

`groovy-json`(version 2.4.4) is a transitive dependency for  `rest-assured` library (via `json-path`) and results in following error when tests are run using Java 11.

`java.lang.ClassCastException: class [B cannot be cast to class [C ([B and [C are in module java.base of loader 'bootstrap')`

This is due to bug (https://bugs.openjdk.java.net/browse/JDK-8218540?focusedCommentId=14244740&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14244740) in `groovy-json` version and is fixed in latest version of the library.

## How to test

- CI should pass
- Tests should run on Java11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -273,6 +273,12 @@
             <version>2.4.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-json</artifactId>
+            <version>2.5.5</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>


pom.xml
```diff
@@ -273,6 +273,11 @@
             <version>2.4.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-json</artifactId>
+            <version>2.5.5</version>
```
danworth: Is this necessary only for tests? Just wondering because I think you mentioned the dependency is via json-path which is scoped for test.
```
<dependency>
            <groupId>com.jayway.jsonpath</groupId>
            <artifactId>json-path-assert</artifactId>
            <version>2.4.0</version>
            <scope>test</scope>
        </dependency>
```

kbottla: Yes, I have just updated it :)

danworth: Just a thought, I see we're using `rest-assured` in its old repo which bundles with groovy 2.4.4 (https://mvnrepository.com/artifact/com.jayway.restassured/rest-assured/2.9.0). It mentioned its' moved to the following location and that version has groovy 2.4.15 https://mvnrepository.com/artifact/io.rest-assured/rest-assured/3.3.0

Does moving the the latest `rest-assured` fix the problem and avoid us having to manage a transient dependency directly in our pom?

kbottla: That's good spot 👍, will have a look.  I guess dependabot can't do these kind of upgrades.

kbottla:
Closing this as upgrading `rest-assured` will be fine https://github.com/alphagov/pay-adminusers/pull/435

PR #435, opened 2019-03-19
PP-4952 Upgrade `REST Assured` library

kbottla:
## WHAT YOU DID

Prep for Java11:

- Upgraded `rest-assured` library which also uses transitive dependency `groovy-json` (version 2.4.15) instead of 2.4.4 which fails on tests for Java 11 (See PR https://github.com/alphagov/pay-adminusers/pull/434 for details)

## How to test
- CI should pass
- Tests should run on Java11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,9 +214,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.jayway.restassured</groupId>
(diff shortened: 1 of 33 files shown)


rhowe-gds:
If this works, we should do the same in cardid, publicapi and publicauth

PR #439, opened 2019-03-26
BAU - add York Uni as they want to use pay 🎓

jonheslop:


46bit:
It's worth noting that this will allow non-staff to sign up. For instance, students also get `york.ac.uk` email addresses. Is that OK?

jonheslop:
yeah we talked about this and decided it’s fine, would be difficult to manage anyway

PR #444, opened 2019-04-02
PP-5061 Migration SQL to backfill English-language service names

alexbishop1:
Database migration to insert a row with the English-language service name into the `service_names` table if there is not already one present for the service.

This should ensure that every English-language service name is duplicated in the `service_names` table.

This won’t lock the table so it’s possible that other service name changes made while this is running may result in inconsistencies and there’s a tiny chance a service trying to update their service
name while this is running may encounter an error. However, the chances are very small (service names are not changed very often) and we can easily check for inconsistencies afterwards and fix any
that are found (which at the very worst may just mean old service names continue to be used for longer than anticipated).

### DIFF
--- a/src/main/resources/migrations/00055_backfill_english_service_names.sql
+++ b/src/main/resources/migrations/00055_backfill_english_service_names.sql
@@ -1,0 +1,11 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:backfill_english_service_names
+INSERT INTO service_names (service_id, language, name)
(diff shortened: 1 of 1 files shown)


alexbishop1:
We already have code that updates the English-language service name in both places when one is updated. It seems to be working (there’s one discrepancy in production but it’s clearly a test service).

Just to be certain, we’ll check that there aren’t any discrepancies before making major changes like only reading from the `service_names` table and stopping updating the `name` in the `services` table.

PR #453, opened 2019-04-05
PP-5601 Remove legacy service name update mechanism

alexbishop1:
Remove support for updating service names by sending a `PATCH` to `/{serviceExternalId}` with the legacy payload:

```json
[
  {
    "op": "replace",
    "path": "name",
    "value": "New Service Name"
  }
]
```

The new payload continues to be supported:

```json
[
  {
    "op": "replace",
    "path": "service_name/en",
    "value": "New Service Name"
  }
]
```

Nothing uses the old payload any more.

Some tests tested updates with the old `"name"` path even though the particular path wasn’t important. These have been updated.


src/main/java/uk/gov/pay/adminusers/model/Service.java
```diff
@@ -19,6 +19,8 @@
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 public class Service {
 
+    public static final String FIELD_NAME = "name";
```
oswaldquek: is there a reason why `FIELD_NAME` is in this class when it's only used by `ServiceResource`?

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
```diff
@@ -143,7 +141,7 @@ public ServiceUpdateOperationValidator(RequestValidations requestValidations) {
     }
 
     private List<String> validateServiceNameValue(JsonNode operation, String path) {
-        boolean allowEmpty = !FIELD_NAME.equals(path) && !path.endsWith('/' + SupportedLanguage.ENGLISH.toString());
+        boolean allowEmpty = !path.endsWith('/' + SupportedLanguage.ENGLISH.toString());
```
oswaldquek: should the `path.endsWith` not check for `SupportedLanguage.WELSH.toString()` as well?

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceUpdateTest.java
```diff
@@ -143,39 +120,12 @@ public void shouldUpdateName_andServiceNameEn_whenReplaceServiceNameEn() {
     }
 
     @Test
-    public void shouldUpdateName_ServiceNameEn_andServiceNameCy_whenReplaceName_AndReplaceServiceNameCy_withNoExistingServiceNameCy() {
+    public void shouldUpdateName_ServiceNameEn_andServiceNameCy_whenReplaceServiceNameEn_AndReplaceServiceNameCy_withNoExistingServiceNameCy() {
```
oswaldquek: this really does my head in. how about `should_update_existing_en_service_name_and_non_existing_cy_service_name`?

src/main/java/uk/gov/pay/adminusers/model/Service.java
alexbishop1: When I moved it, it was used in other places too but those usages have been removed now, so I’ll move the constant to `ServiceResource`.

src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java
alexbishop1: No, the validation rules are subtly different. We don’t allow an empty string to be set as the English-language service name. However, we do allow an empty string to be set as the Welsh-language service name. It’s our way of letting someone remove a Welsh-language service name if they decide they don’t want one any more (there’s logic elsewhere to ignore empty strings for the translated service names when building service objects).

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceUpdateTest.java
alexbishop1: I’ll have a look at the names for all the tests in this class to keep them consistent.

alexbishop1:
Addressed comments:

- Amended first commit to move `FIELD_NAME` from `Service` to `ServiceResource` (no other changes)
- Added another commit to rename tests in `ServiceResourceUpdateTest`

PR #458, opened 2019-04-08
PP-5061 Ignore "name" property of service creation requests

alexbishop1:
When processing a request to create a new service, ignore the `"name"` property and only use the names from the `"service_name"` property.

Only selfservice creates services and it no longer sends the `"name"` property (but it does always include `"service_name"` with at least an `"en"` property).

Since this involves changes to the `ServiceResource` and `ServiceCreator`, make them not use an `Optional<List<String>>` for the gateway account IDs because optionals of lists are cursed.

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceCreateTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceCreateTest.java
@@ -12,7 +12,6 @@ import org.mockito.Captor;
 import org.mockito.Mockito;
 import org.mockito.junit.MockitoJUnitRunner;
 import uk.gov.pay.adminusers.model.Service;
-import uk.gov.pay.adminusers.model.ServiceName;
 import uk.gov.pay.adminusers.persistence.dao.UserDao;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceCreateTest.java
```diff
@@ -200,11 +197,11 @@ public void shouldSuccess_whenCreateAServiceWithName_andGatewayAccountIds() {
     public void shouldSuccess_whenCreateAServiceWithEnglishName_andGatewayAccountIds() {
         PAYLOAD_MAP.put("service_name", Map.of(SupportedLanguage.ENGLISH.toString(), EN_SERVICE_NAME));
         String anotherGatewayAccountId = "another-gateway-account-id";
-        List<String> gatewayAccounts = Arrays.asList(GATEWAY_ACCOUNT_ID, anotherGatewayAccountId);
+        List<String> gatewayAccounts = List.of(GATEWAY_ACCOUNT_ID, anotherGatewayAccountId);
```
stephencdaly: 💯

PR #459, opened 2019-04-08
Remove dependency-check plugin from POM

alexbishop1:
We only ever run this manually, which we can do without it being in the POM:

```
$ mvn org.owasp:dependency-check-maven:check
```


alexbishop1:
We do have things like Synk that monitor our repos for vulnerabilities but we could configure dependency-check to run on a schedule and email the results or something.

PR #464, opened 2019-04-10
Small cleanup to docker-startup.sh

rhowe-gds:
Bash has a more concise way to do environment variable default values. Use it.

### DIFF
--- a/docker-startup.sh
+++ b/docker-startup.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 set -eu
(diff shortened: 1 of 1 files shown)


docker-startup.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 set -eu
-RUN_MIGRATION=${RUN_MIGRATION:-false}
-RUN_APP=${RUN_APP:-true}
+
+: "${RUN_MIGRATION:=false}
```
whpearson: Needs a closing " for both.

PR #534, opened 2019-07-11
REP-137: Remove redundant environment variables

vixus0:
There's no need to have separate environment variables for every single
selfservice URL as there's nowhere in Pay's codebase that makes use of
this functionality. It would also help with overriding the base
selfservice URL for end-to-end tests and we could remove some extra
environment variables from our Terraform code.

## WHAT YOU DID
Removed `SELFSERVICE_*_URL` environment variables from the app's
YAML config and README.

## How to test
Functionality should be unaffected, so make sure everything works as
usual.

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

- [x] Updated README.md for any of the following ?


jonheslop:
Nice one

PR #551, opened 2019-08-12
Small cleanups

rhowe-gds:
## WHAT YOU DID
- Remove the 'ping' healthcheck. It's not needed since we have other healthchecks and all healthchecks are reported together.
- Make some exceptions package-private where they can be
- Define the graphite port number as an `Integer` so that we can offload the parsing of it to Jackson.


src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersConfig.java
```diff
@@ -24,7 +20,7 @@
     @NotNull
     private String graphiteHost;
     @NotNull
-    private String graphitePort;
+    private Integer graphitePort;
```
sfount: 👍

PR #571, opened 2019-09-03
REP: Stop misusing asynchronous calls to Notify

vixus0:
The `NotificationService` was returning things as `CompletableFuture` in
some hopes of making the calls to Notify's API asynchronous. However the
status of the calls was never checked and it looks like adminusers
doesn't reallu care whether the API call was successful or not and
carries on regardless. 

## WHAT YOU DID
I've not changed this behaviour but instead
explicitly catch the `Exception`s thrown by `NotificationService` and
avoided the mess introduced by `CompletableFuture`.

Error logging behaviour should be the same.

## How to test

- Run unit and integration tests and check the build is successful

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

- [x] Updated README.md for any of the following ?

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
+++ b/src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
@@ -98,7 +98,6 @@ public class AdminUsersModule extends AbstractModule {
     @Provides
     private NotificationService provideUserNotificationService() {
         return new NotificationService(
-                environment.lifecycle().executorService("2fa-sms-%d").build(),
                 configuration.getNotifyConfiguration(),
                 configuration.getNotifyDirectDebitConfiguration(),
                 environment.metrics());
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteService.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteService.java
@@ -62,14 +62,15 @@ public class InviteService {
             inviteDao.merge(invite);
             int newPassCode = secondFactorAuthenticator.newPassCode(invite.getOtpKey());
             String passcode = String.format(Locale.ENGLISH, SIX_DIGITS_WITH_LEADING_ZEROS, newPassCode);
-            notificationService.sendSecondFactorPasscodeSms(inviteOtpRequest.getTelephoneNumber(), passcode)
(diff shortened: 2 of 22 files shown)


src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
```diff
@@ -98,7 +98,6 @@ private JpaPersistModule jpaModule(AdminUsersConfig configuration) {
     @Provides
     private NotificationService provideUserNotificationService() {
         return new NotificationService(
-                environment.lifecycle().executorService("2fa-sms-%d").build(),
```
alexbishop1: Presumably we don’t need this at all because email sending is not async any more?

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -62,14 +62,15 @@ public void reGenerateOtp(InviteOtpRequest inviteOtpRequest) {
...
-                    });
+            
             LOGGER.info("New 2FA token generated for invite code [{}]", inviteOtpRequest.getCode());
+            
+            try {
+                String notificationId = notificationService.sendSecondFactorPasscodeSms(inviteOtpRequest.getTelephoneNumber(), passcode);
+                LOGGER.info("sent 2FA token successfully for invite code [{}], notification id [{}]", inviteOtpRequest.getCode(), notificationId);
+            } catch (Exception e) {
+                LOGGER.error(format("error sending 2FA token for invite code [%s]", inviteOtpRequest.getCode()), e);
+            }
```
alexbishop1: Presumably this was reordered to log before sending the email because the previous async approach means the log line would be written before the email sending had finished anyway?

src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
vixus0: Yeah, none of the calls made by `NotificationService` should be async.

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
vixus0: That's right.

PR #589, opened 2019-10-01
Fix PMD warnings from Java “Best Practices” rule set

alexbishop1:
Fix PMD warnings from Java “Best Practices” rule set (except for those rules we exclude — this PR adds a couple of new exclusions).

See individual commits for details.


src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
```diff
@@ -96,7 +96,7 @@ private JpaPersistModule jpaModule(AdminUsersConfig configuration) {
     }
 
     @Provides
-    private NotificationService provideUserNotificationService() {
+    public NotificationService provideUserNotificationService() {
```
oswaldquek: It seems the private keyword on the method didn't stop the app from running...

src/main/java/uk/gov/pay/adminusers/model/PatchRequest.java
```diff
@@ -26,6 +26,10 @@ public static PatchRequest from(JsonNode node) {
         return new PatchRequest(node.get("op").asText(), node.get("path").asText(), node.get("value").asText());
     }
 
+    public String getOp() {
+        return op;
+    }
+
```
oswaldquek: what breakage does adding this getter method fix?

src/main/java/uk/gov/pay/adminusers/app/config/AdminUsersModule.java
alexbishop1: It stops PMD complaining about an apparently unused private method.

src/main/java/uk/gov/pay/adminusers/model/PatchRequest.java
alexbishop1: It stops PMD warning about an unused private field.

PR #592, opened 2019-10-02
PP-5670: Specify the consumer when running contract test

oswaldquek:
Currently when a consumer build is kicked off, the provider contract test will
run in its entirety. For example, in a selfservice build, the adminusers provider
contract tests will run as well. However, what is run is:

products-ui->adminusers
selfservice->adminusers

We really only want to run selfservice->adminusers in this case.

This commit will allow a dynamic creation of the test suite to run based on a
CONSUMER system property. This will mean shorter run times when a provider
contract test is run as part of a specific consumer build.

## WHAT YOU DID
_A brief description of the pull request:_

## How to test

- How should it be reviewed? 
- What feedback do you need? 
- If manual testing is needed, give suggested testing steps.

### Logging

### Documentation

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/pact/ContractTestSuite.java
+++ b/src/test/java/uk/gov/pay/adminusers/pact/ContractTestSuite.java
@@ -1,13 +1,20 @@
 package uk.gov.pay.adminusers.pact;
 
+import junit.framework.JUnit4TestAdapter;
+import junit.framework.TestSuite;
 import org.junit.runner.RunWith;
-import org.junit.runners.Suite;
+import org.junit.runners.AllTests;
+import uk.gov.pay.commons.testing.pact.provider.CreateTestSuite;
 
-@RunWith(Suite.class)
+import java.util.Map;
 
-@Suite.SuiteClasses({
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/adminusers/pact/ContractTestSuite.java
```diff
@@ -1,13 +1,40 @@
...
+            map.forEach((key, value) -> suite.addTest(value));
+        } else if (map.containsKey(consumer)) {
+            LOGGER.info("Running {}-connector contract tests only.", consumer);
+            suite.addTest(map.get(consumer));
+        } else {
+            throw new RuntimeException(format("Error running provider contract tests. ${CONSUMER} system property was %s.", consumer));
+        }
 
+        return suite;
+    }
```
stephencdaly: Reckon we could add a base class or method somewhere in pay-java-commons for this? Might be overkill though..

oswaldquek: Yeah sounds good

PR #603, opened 2019-10-07
BAU Clean up codacy lint errors

danworth:
- Remove unused import
- Remove try block catching a NullPointerException when processing a
`CreateUserRequest` in favour of safely getting each value from the Json payload.

## WHAT YOU DID
I ran PMD locally and fixed two of lint errors.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
@@ -7,7 +7,9 @@ import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.adminusers.app.util.RandomIdGenerator.newId;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
```diff
@@ -42,30 +44,39 @@ public static CreateUserRequest from(String username, String password, String em
...
         }
+
+        var results = new ArrayList<String>();
+        node.get(fieldName)
+                .iterator()
+                .forEachRemaining(nodeValue -> safelyGetNodeAsText(nodeValue).ifPresent(results::add));
+        return results;
+    }
+
+    private static String safelyGetRequiredNodeAsText(JsonNode node, String fieldName) {
```
alexbishop1: There’s nothing really safe about this! I’d call it `getRequiredNodeOrFail` or something.

src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
```diff
@@ -42,30 +44,39 @@ public static CreateUserRequest from(String username, String password, String em
...
+        return safelyGetNodeAsText(node.get(fieldName))
+                .orElseThrow(() -> new RuntimeException(String.format("Error retrieving field {} for creating a user", fieldName)));
+    }
+
+    private static Optional<String> safelyGetNodeAsText(JsonNode node) {
+        return node == null ? Optional.empty() : Optional.ofNullable(node.asText());
```
alexbishop1: This could be `return Optional.ofNullable(node).map(JsonNode::asText)`, which is possibly small enough it could just be inlined inside `safelyGetList(…)`:

```java
// …
forEachRemaining(Optional::ofNullable).map(JsonNode::asText).ifPresent(results::add));
// …
```

src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
danworth: Yeah good point, I named the method then later moved the common safe bit out to be shared with the safelyGetList. I'll update 👍

src/main/java/uk/gov/pay/adminusers/model/CreateUserRequest.java
danworth: Could do, but I was trying to avoid that as I'm not convinced wrapping values we already have in a new optional to avoid a simple null check is a good idea, but then perhaps its a bit more fluid to read and I should get on board with it. I'll try it on and see how it feels :)

PR #605, opened 2019-10-09
BAU Codacy clean up

danworth:
- Change `Link.Rel` enum values to upper case as per convention.

## WHAT YOU DID
Ran PMD locally and fixed the final lint warning.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Link.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Link.java
@@ -5,7 +5,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
-import java.util.Map;
+import java.util.Locale;
 import java.util.Objects;
 
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/model/Link.java
```diff
@@ -37,6 +38,11 @@ private Link(@JsonProperty("rel") Rel rel,
         this.href = href;
     }
 
+    @JsonProperty("rel")
+    public String getRelAsLowerCase() {
+        return rel.name().toLowerCase(Locale.UK);
```
alexbishop1: `Locale.ENGLISH` is probably slightly more appropriate for this purpose (our lower-casing is English-specific but not UK-specific)

danworth: 👍 done.

danworth:
> This is a nice change. Could you check two things:
> 
> 1. If you remove `@JsonProperty("rel")` from `getRelAsLowerCase()`, does an integration test start failing? If not, we should probably add one.
> 2. Can the `@JsonProperty` annotation be removed from the constructor parameters? I think they’re only relevant to deserialising JSON to an object, which I don’t think we do for this.

1. Yes, 12 tests fail, as an example:
```
[ERROR]   ServiceResourceFindTest.shouldGet_existingServiceById_withDefaultEnNameVariant:92->ServiceResourceBaseTest.assertLinks:27 
Expected: is "self"
     but: was "SELF"
```
2. Good point, there's also quite a bit of unused stuff in `Link` which I can remove and make the remainder `final`.

PR #619, opened 2019-10-24
PP-5717: Configure adminusers to use json logging

oswaldquek:
## WHAT YOU DID
_A brief description of the pull request:_

## How to test

- How should it be reviewed? 
- What feedback do you need? 
- If manual testing is needed, give suggested testing steps.

### Logging

### Documentation

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -9,11 +9,11 @@ server:
 logging:
   level: INFO
   appenders:
-    - type: console
+    - type: logstash-console
       threshold: ALL
-      timeZone: UTC
       target: stdout
-      logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [thread=%thread] [level=%-5level] [logger=%logger{15}] [requestID=%X{X-Request-Id:-(none)}] - %msg%n"
+      customFields:
+        container: "adminusers"
     - type: sentry
       threshold: ERROR
       dsn: ${SENTRY_DSN:-https://example.com@dummy/1}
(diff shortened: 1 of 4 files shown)


src/main/resources/config/config.yaml
```diff
@@ -9,11 +9,11 @@ server:
...
   level: INFO
   appenders:
-    - type: console
+    - type: logstash-console
       threshold: ALL
-      timeZone: UTC
       target: stdout
-      logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [thread=%thread] [level=%-5level] [logger=%logger{15}] [requestID=%X{X-Request-Id:-(none)}] - %msg%n"
+      customFields:
+        container: "connector"
```
stephencdaly: should be adminusers

danworth: shouldn't this be 'adminusers'?

src/test/resources/config/test-it-config.yaml
```diff
@@ -9,11 +9,11 @@ server:
...
   level: INFO
   appenders:
-    - type: console
+    - type: logstash-console
       threshold: ALL
-      timeZone: UTC
       target: stdout
-      logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [thread=%thread] [level=%-5level] [logger=%logger{15}] [requestID=%X{X-Request-Id:-(none)}] - %msg%n"
+      customFields:
+        container: "connector"
```
danworth: shouldn't this be `adminusers`?

src/main/resources/config/config.yaml
oswaldquek: whoops!!

oswaldquek: fixed

PR #658, opened 2020-01-02
PP-5313 Inline path strings

stephencdaly:
Inline path string in the UserResource and ForgottenPasswordResource to
make reading the code easier. This now matches what we do in other
resources.


sfount:
🙌 🙌

PR #664, opened 2020-01-08
PP-6001 Upgrade Dropwizard

kbottla:
## WHAT 
- Upgrades dropwizard to v2.0.0
- Added `dropwizard-dependencies` as parent as v2.0.0 upgrade doesn't specify transitive dependency versions anymore https://www.dropwizard.io/en/release-2.0.x/manual/upgrade-notes/upgrade-notes-2_0_x.html

      We can also include BOM for transitive dependency versions but versions can't be overridden and we may end up with multiple revisions of same library. For example, `jackson-version` current used is `2.10.2` but dropwizard uses `2.10.1`

- Upgraded dropwizard-sentry library too without which builds are failing with error `java.lang.NoSuchMethodError: 'com.google.common.collect.ImmutableList org.dhatim.dropwizard.sentry.logging.SentryAppenderFactory.getFilterFactories()'`

- Upgrade pay-java-commons library which is on dropwizard v2.0.0 and defines required library `org.glassfish.jersey.bundles.repackaged`

- Return empty response explictly for /v1/api/invites/user resource as Jersey default response is being returned

    ```
    "servlet":"io.dropwizard.jersey.setup.JerseyServletContainer-6dff6ce",
    "message":"Not Found",
    "url":"/v1/api/invites/user",
    "status":"404"
    ```

- Fixed tests DB setup

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
+++ b/src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
@@ -23,6 +23,7 @@ import static java.lang.String.valueOf;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.RandomUtils.nextInt;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.emptyOrNullString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.isEmptyOrNullString;
@@ -85,7 +86,7 @@ public class UserDaoIT extends DaoTestBase {
         assertThat(userEntity.getId(), is(notNullValue()));
         List<Map<String, Object>> savedUserData = databaseHelper.findUser(userEntity.getId());
         assertThat(savedUserData.size(), is(1));
-        assertThat((String) savedUserData.get(0).get("external_id"), not(isEmptyOrNullString()));
+        assertThat((String) savedUserData.get(0).get("external_id"), not(emptyOrNullString()));
         assertThat(((String) savedUserData.get(0).get("external_id")).length(), equalTo(32));
         assertThat(savedUserData.get(0).get("username"), is(userEntity.getUsername()));
         assertThat(savedUserData.get(0).get("password"), is(userEntity.getPassword()));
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
```diff
@@ -94,8 +95,8 @@ public void shouldCreateAUserSuccessfully() {
         assertThat(savedUserData.get(0).get("telephone_number"), is(userEntity.getTelephoneNumber()));
         assertThat(savedUserData.get(0).get("disabled"), is(Boolean.FALSE));
         assertThat(savedUserData.get(0).get("session_version"), is(0));
-        assertThat(savedUserData.get(0).get("createdAt"), is(java.sql.Timestamp.from(timeNow.toInstant())));
-        assertThat(savedUserData.get(0).get("updatedAt"), is(java.sql.Timestamp.from(timeNow.toInstant())));
+        assertThat(savedUserData.get(0).get("createdat"), is(java.sql.Timestamp.from(timeNow.toInstant())));
+        assertThat(savedUserData.get(0).get("updatedat"), is(java.sql.Timestamp.from(timeNow.toInstant())));
```
jankowiakmaria: Why has it changed?

src/test/java/uk/gov/pay/adminusers/utils/DatabaseTestHelper.java
```diff
@@ -206,14 +207,14 @@ public DatabaseTestHelper addService(Service service, String... gatewayAccountId
...
-            return handle.createStatement("INSERT INTO services(" +
+            return handle.createUpdate("INSERT INTO services(" +
                     "id, custom_branding, " +
                     "merchant_name, merchant_telephone_number, merchant_address_line1, merchant_address_line2, merchant_address_city, " +
                     "merchant_address_postcode, merchant_address_country, merchant_email, external_id) " +
                     "VALUES (:id, :customBranding, :merchantName, :merchantTelephoneNumber, :merchantAddressLine1, :merchantAddressLine2, " +
                     ":merchantAddressCity, :merchantAddressPostcode, :merchantAddressCountry, :merchantEmail, :externalId)")
                     .bind("id", service.getId())
-                    .bind("customBranding", customBranding)
+                    .bindBySqlType("customBranding", customBranding, OTHER)
```
jankowiakmaria: I looked into https://jdbi.org/#_binding_arguments
I'm just wondering whether wouldn't it be clearer to have something like (copied from docs):
```java
QualifiedType<String> json = QualifiedType.of(String.class).with(Json.class);
...
.bindByType("customBranding", customBranding, json);
```

src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
kbottla: JDBI3 doesn't do exact mapping as column names in select statement. Haven't looked into details though

src/test/java/uk/gov/pay/adminusers/utils/DatabaseTestHelper.java
kbottla: We will have to create new JsonArgumentFactory in order to use qualified type and handle string to json conversion within factory. Agree that it looks clearer but it is more setup for tests

```
@Json
public class JsonArgumentFactory extends AbstractArgumentFactory<String> {
    @Override
    protected Argument build(String value, ConfigRegistry config) {
        // do something specifically for json data
    }
}
```

PR #666, opened 2020-01-09
PP-6000 Allow choosing the Notify template for sending OTP via SMS

alexbishop1:
Change `NotificationService` so that when the `sendSecondFactorPasscodeSms(…)` method is called, an enum constant from the new `OtpNotifySmsTemplateId` enum type is passed in to indicate which Notify template ID to use.

For now, all callers of this method pass in the `LEGACY` enum constant, which maps to the long-serving Notify template ID, so there’s no behaviour change yet.

In subsequent commits, different templates will be chosen for different types of messages and `LEGACY` will be removed.

Also add unit tests for `NotificationService` because there weren’t any and the logic is becoming a bit more complex now. This required some tweaks to how `NotificationService` is instantiated to make it easier to inject mock dependencies.

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/service/NotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/adminusers/service/NotificationServiceTest.java
@@ -1,0 +1,119 @@
+package uk.gov.pay.adminusers.service;
+
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.MetricRegistry;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import uk.gov.pay.adminusers.app.config.NotifyConfiguration;
+import uk.gov.pay.adminusers.app.config.NotifyDirectDebitConfiguration;
+import uk.gov.service.notify.NotificationClient;
+import uk.gov.service.notify.NotificationClientException;
+import uk.gov.service.notify.SendSmsResponse;
+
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/adminusers/service/NotificationServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Test
+    public void sendSecondFactorPasscodeSmsWithSignInTemplate() throws NotificationClientException {
+        given(mockNotificationClient.sendSms(SIGN_IN_OTP_SMS_TEMPLATE_ID, PHONE_NUMBER_E164, Map.of("code", OTP), null))
+                .willReturn(mockSendSmsResponse);
+
+        String notificationId = notificationService.sendSecondFactorPasscodeSms(PHONE_NUMBER, OTP, OtpNotifySmsTemplateId.SIGN_IN);
+        
+        assertThat(notificationId, is(NOTIFICATION_ID.toString()));
+    }
```
oswaldquek: I think the real test should be verifying that `mockNotificationClient` is called with the correct template id, phone number and code. I can see this is effectively tested here but in a more long winded way.
It doesn't make sense to assert the `notificationId` (line 86).

alexbishop1: Yeah, I wasn’t too happy with this too. The fact that the the `NotificationClient` returns something pushed me towards structuring the test in a way that hides what’s really important. I’ll change it to verify the call to the `NotificationClient` and not worry about the return value.

PR #669, opened 2020-01-10
PP-6000 Make InviteEntity type field an InviteType not String

alexbishop1:
Make the `InviteEntity` type field an `InviteType` and not a `String.` Also make its the `InviteEntity` `getType()` method return an `InviteType` not a `String`. Use `InviteEntity`’s existing `isUserInvite()` and `isServiceInvite()` methods rather than `getType()` in some places.

This means that any `InviteEntity` that is loaded from the database and has a type that isn’t “user” or “service” will blow up during deserialisation but there are no such entities in the production database and other things would probably break if there were.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
@@ -68,7 +68,8 @@ public class InviteEntity extends AbstractEntity {
     private Integer loginCounter = 0;
 
     @Column(name = "type")
-    private String type = USER.getType();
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteEntity.java
```diff
@@ -201,20 +202,20 @@ public void setLoginCounter(Integer loginCount) {
         this.loginCounter = loginCount;
     }
 
-    public String getType() {
+    public InviteType getType() {
```
sfount: Revolutionary! 👏

PR #675, opened 2020-01-16
PP-5920 Update RDS Cert

danworth:
WHAT YOU DID
Updated the the root CA certificate in the container build in preparation for updating the certs installed on the RDS instances.

The combined cert bundle URL does not need to be updated as the URL has not changed, but the updated bundle will be downloaded at build time.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

How to test
Rebuild containers and deploy to test environment
Ensure app works with NEW CA root cert but OLD RDS certs. The new ca certs must be backward compatible.
Schedule the cert update on the RDS instance.

See similar PR for connector
https://github.com/alphagov/pay-connector/pull/2026

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ENV LANG C.UTF-8
 RUN echo networkaddress.cache.ttl=$DNS_TTL >> "$JAVA_HOME/conf/security/java.security"
 
 # Add RDS CA certificates to the default truststore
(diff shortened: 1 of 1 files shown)


richardTowers:
Do we need to leave the old cert in for backwards compat?

rjbaker:
The updated bundle should include the old cert:

> The new certificate (CA-2019) is available as part of a certificate bundle that also includes the old certificate (CA-2015) so that you can make a smooth transition without getting into a chicken and egg situation.

https://aws.amazon.com/blogs/aws/urgent-important-rotate-your-amazon-rds-aurora-and-documentdb-certificates/

We've tested one app with new certs, old RDS and it works okay.

PR #699, opened 2020-02-21
PP-6152 Service entity experimental features flag

sfount:
Add accessors for the `experimental_features_enabled` services database column.

* Update Entity 
* Update POJO used for resource response 
* Assert values are correctly set and parsed through to JSON

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Service.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Service.java
@@ -30,6 +30,7 @@ public class Service {
     private boolean redirectToServiceImmediatelyOnTerminalState;
     private boolean collectBillingAddress;
     private GoLiveStage goLiveStage;
+    private boolean experimentalFeaturesEnabled;
 
     @JsonIgnore
     private ServiceName serviceName;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/adminusers/model/Service.java
```diff
@@ -30,6 +30,7 @@
     private boolean redirectToServiceImmediatelyOnTerminalState;
     private boolean collectBillingAddress;
     private GoLiveStage goLiveStage;
+    private boolean experimentalFeaturesEnabled;
```
kbottla: Will it for all experimental features ? as the name implies. It is possible that different services try different experimental features at a time and we might accidentally enable for the services not needed.

rorymalcolm: in the ticket we talk about potential extension and migration to a JSONB field I think?

sfount: Yeah as described in https://github.com/alphagov/pay-adminusers/pull/698 this is just a generic "opt in" to whatever the developers are up to -- if we need more permanent control over a feature (like redirect on failure) it should be its own flag. Any features that enable themselves with this flag should be temporary and eventually open to all users. 

If we need more fine grained control over temporary features then we could use JSONB or something as suggested.

kbottla: As discussed (offline), we will be fine with this for current feature. Devs should be aware of the impact of enabling this feature and when adding a new feature, if another feature is in experimental stage or being released to services in phases. We might need to  track the services with the flag set so product can look back.

PR #710, opened 2020-03-03
BAU Edit Email Validation

rorymalcolm:
- Allow a Northern Irish district council to use Pay.


rorymalcolm:
> Please also update the unit tests for this.

Done, forgot we had tests for each individual domain.

PR #718, opened 2020-03-16
PP-6296 Whitelist nmni domain

sfount:
Allow users to sign up with `@nmni.com` emails as per PP-6926.


rorymalcolm:
actually wait, we have tests for this

src/main/java/uk/gov/pay/adminusers/utils/email/EmailValidator.java
```diff
@@ -44,7 +44,8 @@
             "socialworkengland\\.org\\.uk",
             "ucds\\.email",
             "wmca\\.org\\.uk",
-            "york\\.ac\\.uk"
+            "york\\.ac\\.uk",
+            "nmni\\.com"
```
stephencdaly: the domains are in alphabetical order so far, so would be good to keep it that way

sfount:
Thanks @rorymalcolm @stephencdaly, I've moved it to be alphabetically sorted -- testing every single input into a method feels like we're testing Java `util.List` but I can add it if it's necessary

rorymalcolm:
> Thanks @rorymalcolm @stephencdaly, I've moved it to be alphabetically sorted -- testing every single input into a method feels like we're testing Java `util.List` but I can add it if it's necessary

I agree but I had stuff knocked back on this before for not having them

sfount:
Understood @rorymalcolm, added them 👍 https://github.com/alphagov/pay-adminusers/commit/f105e6a9b6845a7660c9f534fb40d67369088204

PR #775, opened 2020-06-18
Remove Apache commons collections

rhowe:
Commons collections is only used in one place, and can be replaced by `List.equals()`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,11 +158,6 @@
             <version>3.10</version>
         </dependency>
         <dependency>
-            <groupId>commons-collections</groupId>
-            <artifactId>commons-collections</artifactId>
-            <version>3.2.2</version>
-        </dependency>
-        <dependency>
             <groupId>org.mindrot</groupId>
             <artifactId>jbcrypt</artifactId>
             <version>0.4</version>
@@ -186,12 +181,6 @@
             <groupId>commons-validator</groupId>
             <artifactId>commons-validator</artifactId>
             <version>1.6</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-collections</groupId>
-                    <artifactId>commons-collections</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.googlecode.libphonenumber</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -186,12 +181,6 @@
             <groupId>commons-validator</groupId>
             <artifactId>commons-validator</artifactId>
             <version>1.6</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-collections</groupId>
-                    <artifactId>commons-collections</artifactId>
-                </exclusion>
-            </exclusions>
```
alexbishop1, line 194: I was wondering if perhaps one reason for excluding commons-validator’s version of commons-collections might have been that it was on an older version with security vulnerabilities but it seems commons-validator 1.6 depends on the same 3.2.2 version of commons-collections as we were already using, so nothing changes there.

src/main/java/uk/gov/pay/adminusers/model/Role.java
```diff
@@ -73,7 +71,7 @@ public boolean equals(Object o) {
             return false;
         }
 
-        return isEqualCollection(permissions, role.permissions);
+        return permissions == null && role.permissions == null || permissions.equals(role.permissions);
```
alexbishop1, line 74: This could actually be `return Objects.equals(permissions, role.permissions);`, which takes care of the nulls. However, it doesn’t look like either argument could actually be null: in fact, `isEqualCollection(…)` will throw a `NullPointerException` is either of its arguments are null, so we can just do `return permissions.equals(role.permissions);`.

However, using `List`’s `equals(…)` method would mean we’re relying on the order of the elements in the two lists being the same. Now they might well be but I’m not sure.

Probably the best approach would be to use a `Set` rather than a `List` to hold collections of permissions, since  it makes no sense to have a permission more than once. `Set`’s `equals(…)` method doesn’t care about order, which would allow us to `return permissions.equals(role.permissions);`. This doesn’t look like a huge change, so I’ll see about doing that today.

alexbishop1:
Closing in favour of https://github.com/alphagov/pay-adminusers/pull/778.

PR #776, opened 2020-06-18
Replace Apache commons-lang with commons-lang3

rhowe:
commons-lang isn't actually declared as a dependency so it's a bit naughty to rely on it being there. Switch to commons-lang3, which we do depend on.


alexbishop1:
Closing in favour of https://github.com/alphagov/pay-adminusers/pull/779.

PR #780, opened 2020-06-20
Remove 'migrateToInitialDbState' command

rhowe:
We can use liquibase preconditions to only apply these migrations if they're
needed and we can use `onFail="MARK_RAN"` to update the `databasechangelog` table
such that they won't get run again.

This puts the entire database state under the control of `migrations.xml` and the
primary `db migrate` command from `dropwizard-migrations`, simplifying things
substantially.

Now that `dropwizard-migrations` is our prime mechanism to set the database up, use it when setting things up for tests, allowing us to remove our Liquibase dependency. This does downgrade Liquibase slightly to the version included in `dropwizard-migrations` but that's OK.

I don't think this change will lead to the production database getting eaten, but _caveat emptor_.


rhowe:
Removing the command entirely might be a bit much - could leave a no-op one behind

rhowe:
This needs a bit of a rework following the update to junit 5. Closing.

PR #781, opened 2020-06-22
PP-6680 Include tag of Docker base image in FROM line in Dockerfile

alexbishop1:
This should help Dependabot to automatically find updates.


rhowe:
Not supposed to be necessary since https://github.com/dependabot/dependabot-core/pull/100 but :shrug:

rjbaker:
Dependabot can't resolve the tag from only the digest, so we never get any Dependabot updates.  See: https://github.com/dependabot/feedback/issues/129
This change allows Dependabot to resolve the tag, thus any new changes whilst still allowing us to pin the digest. Also, hi 👋 @rhowe 😄

rhowe:
:wave: 
sounds like the GDS Way Dockerfile guidance could do with an update then - blame the person that wrote it for not testing their assumptions!

alexbishop1:
Even if it wasn’t necessary, I like it because it makes it more obvious what the Docker image has in it.

PR #784, opened 2020-06-28
Make RoleDaoIT deterministic

rhowe:
Roles are immutable, so we can tie the DAO test to the preloaded contents of
the database rather than generating random ones for testing against.

Also, we are only testing role retrieval, not retrieval of associated
permissions so we don't need the extra stuff RoleDbFixture does.

Add a test for the behaviour when the requested role does not exist.


katstevens:
Closing this as it's gone stale.

PR #786, opened 2020-06-29
PP-6257 Refactor service update validator tests

stephencdaly:
Make tests easier to read by using parameterized tests.
Add tests for "gateway_account_id" path validation.
Make validation message consistent when value is null or missing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -281,6 +281,12 @@
             <artifactId>jaxb-api</artifactId>
             <version>2.3.1</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


stephencdaly:
> Nice consolidation of the tests. It's tricky that we are moving towards Junit5 and JunitParams doesn't have support for Junit5, but should be good for now.

Oh yeh, hopefully whatever implementation works for JUnit5 is not too hard to migrate from this.
It looks like it works similarly https://www.baeldung.com/parameterized-tests-junit-5#6-method

PR #789, opened 2020-06-30
PP-6257 Return new service fields in service response

stephencdaly:
Read new fields from the database and return them in the get service response.


src/main/java/uk/gov/pay/adminusers/model/Service.java
```diff
@@ -31,6 +34,15 @@
     private boolean collectBillingAddress;
     private GoLiveStage goLiveStage;
     private boolean experimentalFeaturesEnabled;
+    private String sector;
```
alexbishop1, line 37: What values can `sector` have? Might it be a good candidate for an enum?

stephencdaly: Thoughts were that because it's only for reporting we'll only constrain this in toolbox.
Makes it easier to add to or change when we need to

stephencdaly: Plan is to have a drop-down in toolbox of allowed values

PR #798, opened 2020-07-03
PP-6708 Upgrade unit tests to JUnit5

kbottla:


src/test/java/uk/gov/pay/adminusers/utils/DateTimeUtilsTest.java
```diff
@@ -1,13 +1,12 @@
 package uk.gov.pay.adminusers.utils;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 import java.time.ZonedDateTime;
 
-import static junit.framework.TestCase.assertEquals;
```
alexbishop1, line 7: How did we ever have this? `junit.framework` went away in JUnit 4, which came out in 2006, 9 years before the Payments Platform project (as it was then known) started and 11 years before the first in of adminusers code was written!

kbottla: 🤷‍♂️

PR #799, opened 2020-07-05
Remove misuse of Liquibase exception

rhowe:
We shouldn't be throwing one of Liquibase's exceptions in our own code.

Split out from https://github.com/alphagov/pay-adminusers/pull/780


rhowe:
This was fixed in a3763bd63ddc10871fa150d78e891f9ce6a5fe30

PR #830, opened 2020-08-07
PP-6883 Add Template ID

rorymalcolm:
- Add the template ID for the Notify invite to service email


src/main/resources/env-map.yml
```diff
@@ -1,14 +1,15 @@
...
+  SELFSERVICE_URL:                         '.[][] | select(.name == "app-catalog")               | .credentials.selfservice_url'
+  SUPPORT_URL:                             '.[][] | select(.name == "app-catalog")               | .credentials.card_frontend_support_url'
+  DB_HOST:                                 '.[][] | select(.name == "adminusers-db")             | .credentials.host'
+  DB_NAME:                                 '.[][] | select(.name == "adminusers-db")             | .credentials.name'
+  DB_PASSWORD:                             '.[][] | select(.name == "adminusers-db")             | .credentials.password'
+  DB_USER:                                 '.[][] | select(.name == "adminusers-db")             | .credentials.username'
+  DB_SSL_OPTION:                           '.[][] | select(.name == "adminusers-db")             | .credentials.ssl_option // "ssl=true"'
+  NOTIFY_API_KEY:                          '.[][] | select(.name == "adminusers-secret-service") | .credentials.notify_api_key'
+  NOTIFY_INVITE_SERVICE_EMAIL_TEMPLATE_ID: '.[][] | select(.name == "adminusers-secret-service") | .credentials.notify_invite_service_email_template_id'
```
danworth, line 11: do these template ids need to change per environment or is it always the same? If so it could go into the manifest directly.

rorymalcolm: largely the same but there are some which swap around based on environment

PR #849, opened 2020-08-26
PP-7011 Adding MOTO masking permissions

narinderchana:
Description:
- Adding permissions for reading and updating MOTO mask input settings

### DIFF
--- a/src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
+++ b/src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
@@ -1,0 +1,14 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:moto_mask_input_update_permission
+INSERT INTO permissions(id, name, description) VALUES (49, 'moto-mask-input:update', 'Update MOTO mask input settings');
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:update'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:update'), NOW(), NOW());
+
+--changeset uk.gov.pay:moto_mask_input_read_permission
+INSERT INTO  permissions(id, name, description) VALUES (50, 'moto-mask-input:read', 'View MOTO mask input settings');
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:read'), NOW(), NOW());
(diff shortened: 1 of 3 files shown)


src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
```diff
@@ -0,0 +1,14 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:moto_masking_account_update_permission
```
alexbishop1, line 3: Changeset ID doesn’t quite match the permission: should be `uk.gov.pay:moto_mask_input_update_permission`.

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO permissions(id, name, description) VALUES (49, 'moto-mask-input:update', 'Update Moto Masking Account');
```
alexbishop1, line 4: The description should be something like “Update MOTO mask input settings”.

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:update'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'moto-mask-input:update'), NOW(), NOW());
+
+--changeset uk.gov.pay:moto_masking_account_read_permission
```
alexbishop1, line 8: Changeset ID doesn’t quite match the permission: should be `uk.gov.pay:moto_mask_input_read_permission`.

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO  permissions(id, name, description) VALUES (50, 'moto-mask-input:read', 'View Moto Masking Account');
```
alexbishop1, line 9: The description should be something like “View MOTO mask input settings”.

alexbishop1 (changes requested):
Looks fine apart from a few description strings.

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
narinderchana: ok will make that change

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
narinderchana: ok will make that change

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
narinderchana: ok will make that change

src/main/resources/migrations/00063_add_moto_account_masking_permisions.sql
narinderchana: ok will make that change

alexbishop1 (approved):

PR #855, opened 2020-09-04
PP-6913 Change Env Map

rorymalcolm:
- Bring the env vars from the new correct place


src/main/resources/env-map.yml
```diff
@@ -2,10 +2,10 @@ env_vars:
...
-  DB_PASSWORD:                                                                  '.[][] | select(.name == "adminusers-db")             | .credentials.password'
-  DB_USER:                                                                      '.[][] | select(.name == "adminusers-db")             | .credentials.username'
+  DB_HOST:                                                                      '.[][] | select(.name == "adminusers-secret-service") | .credentials.db_host'
+  DB_NAME:                                                                      '.[][] | select(.name == "adminusers-secret-service") | .credentials.db_name'
+  DB_PASSWORD:                                                                  '.[][] | select(.name == "adminusers-secret-service") | .credentials.db_password'
+  DB_USER:                                                                      '.[][] | select(.name == "adminusers-secret-service") | .credentials.db_user'
   DB_SSL_OPTION:                                                                '.[][] | select(.name == "adminusers-db")             | .credentials.ssl_option // "ssl=true"'
```
rjbaker, line 9: This can probably be removed or added to the static values as `adminusers-db` will no longer exist. Or we can set `DB_SSL_OPTION: true` in the app manifest.

rjbaker (changes requested):

rjbaker (approved):
Looks good 👍

PR #917, opened 2020-12-22
bau: Reduce ObjectMapper instantiation

oswaldquek:
ObjectMapper is an expensive object so reduce its instantiation.
Also remove [some] unnecessary `public` qualifiers not needed for junit5 tests.


sfount (approved):
Nice one 👍

PR #926, opened 2021-01-12
zd4422632: Allow email from Fermanagh & Omagh District Council

oswaldquek:


src/main/java/uk/gov/pay/adminusers/utils/email/EmailValidator.java
```diff
@@ -56,7 +56,8 @@
             "ucds.email",
             "wmca.org.uk",
             "york.ac.uk",
-            "digitalaccessibilitycentre.org"
+            "digitalaccessibilitycentre.org",
+            "fermanaghomagh.com"
```
alexbishop1, line 60: Can you put this in the right place alphabetically please? (It’s generally in alphabetical order, except for digitalaccessibilitycentre.org, which I assume is because they were given access for the accessibility audit rather than because it’s a public sector domain.)

src/test/java/uk/gov/pay/adminusers/utils/email/EmailValidatorIsPublicSectorEmailTest.java
```diff
@@ -129,13 +129,14 @@
                 {"valid@subdomain.wmca.org.uk", true},
                 {"valid@subdomain.york.ac.uk", true},
                 {"valid@subdomain.digitalaccessibilitycentre.org", true},
+                {"valid@subdomain.fermanaghomagh.com", true},
```
alexbishop1, line 132: There should be too of these: `valid@fermanaghomagh.com` (in this group higher up) and `valid@subdomain.fermanaghomagh.com` (and, again, in alphabetical order).

alexbishop1 (changes requested):

alexbishop1 (approved):

PR #932, opened 2021-01-20
PP-7663: Endpoint to get admin emails for gateway accounts

oswaldquek:


src/test/java/uk/gov/pay/adminusers/resources/UserResourceGetIT.java
```diff
@@ -18,10 +19,52 @@
...
+        
+        Role viewOnlyRole = roleDbFixture(databaseHelper).withName("view-only").insertRole();
+        userDbFixture(databaseHelper).withServiceRole(service.getId(), viewOnlyRole.getId()).insertUser();
+
+        givenSetup()
+                .when()
+                .contentType(JSON)
+                .accept(JSON)
+                .queryParam("gatewayAccountId", gatewayAccount1)
+                .queryParam("gatewayAccountId", gatewayAccount2)
```
oswaldquek, line 56: The request will look something like `/v1/api/users/admin-emails-for-gateway-accounts?gatewayAccountId=206979493&gatewayAccountId=429850515`

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
```diff
@@ -90,6 +90,15 @@ public Response getUser(@PathParam("userExternalId") String externalId) {
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
     
+    @GET
+    @Path("/admin-emails-for-gateway-accounts")
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Map<String, List<String>> getAdminUserEmailsForGatewayAccountIds(
+            @QueryParam("gatewayAccountId") List<String> gatewayAccountIds) {
```
sfount, line 98: Question on this param being a query string, I believe the default serialisation for this is `gatewayAccountId=xxx&gatewayAccountId-xxx`. 

If the number of unique searchable ids could be ~500, would we run into issues with maximum query string policies enforced anywhere along the app/ load balancer/ app proxies? Potentially using a `POST` body would get around this providing almost any data size with multiple body pages.

src/main/java/uk/gov/pay/adminusers/service/UserServices.java
```diff
@@ -126,8 +131,21 @@ public UserServices(UserDao userDao,
...
     }
-
-
+    
+    public Map<String, List<String>> getAdminUserEmailsForGatewayAccountIds(List<String> gatewayAccountIds) {
+        return gatewayAccountIds.stream().collect(Collectors.toMap(
+                gatewayAccountId -> gatewayAccountId,
+                gatewayAccountId -> {
+                    Optional<List<String>> adminUserEmails = serviceFinder.byGatewayAccountId(gatewayAccountId)
+                            .map(service -> userDao.findByServiceId(service.getId())
```
sfount, line 140: This looks like a reasonable first pass to me, I wonder if given the previous example of 500 we'd want to lean on the database to do this rather than iterating through and piecing it together on the app side, I think it would be n gateway accounts * 2 database queries. This likely wouldn't be an issue for the 500 size but might not be resilient to scaling much beyond that.

sfount (commented):
This looks like exactly the functionality required by need and no issues with the code. Some questions on the method which don't have to be blocking

src/main/java/uk/gov/pay/adminusers/resources/UserResource.java
oswaldquek: These are good reasons to switch to using POST with a body. Fixed.

sfount (approved):
Fulfills the requirements of the ticket and the code looks good 👍  As mentioned it may be good to revisit the method for looking up email addresses over to the DAO layer to make it resilient to large requests as more accounts and users are added (this is specifically problematic if JPA optimistically loads all of the services for each of the users, if this is lazy loaded it shouldn't be as much of an impact)

PR #940, opened 2021-02-08
PP-7708 Database migrations for agent-initiated MOTO payments

alexbishop1:
- Add `agent_initiated_moto_enabled` boolean column to services table with default value of false
- Add `agent-initiated-moto:create` permission, granting it to the `admin` role
- Add new `view-and-initiate-moto` role, giving it the same permissions as `view-only` and also `agent-initiated-moto:create`
- Add new `view-refund-and-initiate-moto` role, giving it the same permissions as `view-and-refund` and also `agent-initiated-moto:create`

Tested locally and it did what I expected.


kbottla (approved):
LGTM!

PR #942, opened 2021-02-09
PP-7708 Add agent_initiated_moto_enabled property to service

alexbishop1:
Add `agent_initiated_moto_enabled` property to service, making it updatable in the same way as other boolean properties on service.


sfount (approved):
Code looks good to me, no need for discussion to stop dev progress.

Question raised about the long term effects of having separate moto feature flags on both the gateway account and the service entities. 

It's likely that in the future that all gateway accounts that have been approved for moto (based on PCI) will have access to this feature, it might be worth having a plan for this separate feature flag on the service to be removed or consolidated into the existing gateway account flag.

PR #949, opened 2021-02-15
PP-7765 add new permission

SandorArpa:
## WHAT YOU DID
- add a new permission and role_permissions for psp test account request

### DIFF
--- a/src/main/resources/migrations/00068_add_psp_test_account_stage_permission.sql
+++ b/src/main/resources/migrations/00068_add_psp_test_account_stage_permission.sql
@@ -1,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:psp_test_account_stage_update_permission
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00068_add_psp_test_account_stage_permission.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:psp_test_account_stage_update_permission
+INSERT INTO permissions(id, name, description) VALUES (52, 'psp_test_account_stage:update', 'Update PSP Test Account stage');
```
kbottla, line 4: Can we use Max(ID)+1 for id similar to https://github.com/alphagov/pay-adminusers/blob/221a984b930e679ea983a6fe043e7766ce7f268a/src/main/resources/migrations/00065_add_agent_initiated_moto_create_permission.sql#L4 so we can avoid hard coding. I understand we have mix of both but going towards max(id) will make it less error prone

kbottla (commented):
Looks good to me. Just added a suggestion to use max() to set ID as we already started using it and we no longer need to hardcode IDs.

kbottla (approved):

PR #972, opened 2021-04-07
PP-7964 Replace Collectors.toMap() with Collectors.toUnmodifiableMap()

alexbishop1:
Replace `Collectors.toMap()` with `Collectors.toUnmodifiableMap()`.

The returned maps are never modified and will never contain nulls.


nimalank7 (approved):
LGTM!

PR #980, opened 2021-04-08
BAU Add dropwizard-json-logging dependency

stephencdaly:
pay-java-commons has a dependency on dropwizard-json-logging. Because pay-java-commons uses its own version of dropwizard, the pay-java-commons version and dropwizard dependency versions in adminusers need to be updated at the same time to build successfully.

Add a direct dependency on dropwizard-json-logging so we ignore the version in pay-java-commons and use the version consistent with all the other dropwizard modules.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,11 @@
         </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>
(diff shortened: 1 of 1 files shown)


alexbishop1 (approved):
Sounds sensible.

PR #1012, opened 2021-05-28
PP-7920 Adminusers Drop DB Migration

rorymalcolm:
- Drop a real DB migration for adminusers


katstevens (approved):
Cool - I'll pin the Concourse release to make sure this `drop` doesn't undo the `create` before we've tested it.

PR #1039, opened 2021-07-09
BAU Add the domain `tfgm.com` to the allow list

iqbalgds:
- This will allow users with email addresses ending with tfgm.com to create
  test accounts.


SandorArpa (approved):
LOKTM

PR #1086, opened 2021-08-25
PP-8445 Add default_billing_address_country to services

stephencdaly:
Add column default_billing_address_country to services table and default to "GB". At this stage the only expected values would be "GB" or null.


SandorArpa (approved):
LOKTM

PR #1110, opened 2021-09-29
PP-8728 InviteRequestValidator: remove deprecated method

alexbishop1:
Remove a deprecated method we’re not using an actually unit-test the method we are using.


SandorArpa (dismissed):
LGTM

SandorArpa (approved):
LOKTM

PR #1112, opened 2021-09-29
PP-8728 Set phone and password on service invite when sending OTP

alexbishop1:
If a request is sent to `/v1/api/invites/{code}/otp/generate` with the following body:

```json
{
  "telephone_number": "07700900000",
  "password": "password123"
}
```

… and the code relates to a service invite (that is, one where the user is signing up on their own initiative without receiving an invitation to join a service) then update the phone number and password on the invite with the provided values before sending the one-time password.

This makes `ServiceOtpDispatcher` behave similarly to `UserOtpDispatcher` (the equivalent for user invites — that is, where
a user is signing up because they have been invited to a service) except the `"telephone_number"` and `"password"` properties are optional for `ServiceOtpDispatcher`.

This prepares us for a world where service invites only have the phone number and password set on them immediately before the one-time password is sent rather than when they’re created.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceOtpDispatcher.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceOtpDispatcher.java
@@ -4,8 +4,10 @@ import com.google.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.adminusers.persistence.dao.InviteDao;
+import uk.gov.pay.adminusers.utils.telephonenumber.TelephoneNumberUtility;
 
 import java.util.Locale;
+import java.util.Optional;
 
 import static java.lang.String.format;
 import static uk.gov.pay.adminusers.service.NotificationService.OtpNotifySmsTemplateId.SELF_INITIATED_CREATE_NEW_USER_AND_SERVICE;
@@ -15,13 +17,16 @@ public class ServiceOtpDispatcher extends InviteOtpDispatcher {
     private static final Logger LOGGER = LoggerFactory.getLogger(ServiceOtpDispatcher.class);
     private final InviteDao inviteDao;
     private final SecondFactorAuthenticator secondFactorAuthenticator;
+    private final PasswordHasher passwordHasher;
     private final NotificationService notificationService;
 
     @Inject
-    public ServiceOtpDispatcher(InviteDao inviteDao, SecondFactorAuthenticator secondFactorAuthenticator, NotificationService notificationService) {
+    public ServiceOtpDispatcher(InviteDao inviteDao, SecondFactorAuthenticator secondFactorAuthenticator,
+                                PasswordHasher passwordHasher, NotificationService notificationService) {
         super();
         this.inviteDao = inviteDao;
         this.secondFactorAuthenticator = secondFactorAuthenticator;
+        this.passwordHasher = passwordHasher;
         this.notificationService = notificationService;
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/service/ServiceOtpDispatcher.java
```diff
@@ -15,20 +17,33 @@
...
+                            .map(TelephoneNumberUtility::formatToE164)
+                            .ifPresent(inviteEntity::setTelephoneNumber);
+
+                    Optional.ofNullable(inviteOtpRequest.getPassword())
+                            .map(passwordHasher::hash)
+                            .ifPresent(inviteEntity::setPassword);
+
+                    inviteDao.merge(inviteEntity);
+
                     int newPassCode = secondFactorAuthenticator.newPassCode(inviteEntity.getOtpKey());
```
mrlumbu, line 47: might be worth extracting this whole lambda? It's getting a bit straggly. Happy for that to be in follow up, or for you to tell me otherwise...

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceGenerateOtpIT.java
```diff
@@ -81,6 +80,36 @@ public void generateOtp_shouldSucceed_forServiceInvite_evenWhenTokenIsExpired_si
...
+                "password", PASSWORD);
+
+        givenSetup()
+                .when()
+                .body(mapper.writeValueAsString(invitationRequest))
+                .contentType(ContentType.JSON)
+                .post(format(INVITES_GENERATE_OTP_RESOURCE_URL, code))
+                .then()
+                .statusCode(OK.getStatusCode());
+    }
```
mrlumbu, line 111: It does seem a bit strange to be sending a tel no and password as part of a otp generation request, but without more context I don't know what the other options are, so I trust this is sensible

src/test/java/uk/gov/pay/adminusers/service/ServiceOtpDispatcherTest.java
```diff
@@ -52,6 +64,59 @@ public void shouldSuccess_whenDispatchServiceOtp_ifInviteEntityExist() {
         assertThat(dispatched,is(true));
     }
 
+    @Test
+    public void shouldSuccess_whenDispatchServiceOtp_ifInviteEntityExist_butPhoneAndPasswordOnlyInRequest_andUpdateInviteEntityWithPhoneAndPassword() {
```
mrlumbu, line 68:
```suggestion
    public void shouldSucceed_whenDispatchServiceOtp_ifInviteEntityExist_butPhoneAndPasswordOnlyInRequest_andUpdateInviteEntityWithPhoneAndPassword() {
```
Minorest of minors...

mrlumbu (approved):
Some minor comments, but looks good 👍

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceGenerateOtpIT.java
alexbishop1: It is a bit strange but it’s consistent with how the `UserOtpDispatcher` works. It’s kind of reasonable to set the phone number when sending the OTP but the fact the password is in there is just because the behaviour on adminsers is very tied to the flow in selfservice. But we find the world as it is, not as we would like it to be.

src/test/java/uk/gov/pay/adminusers/service/ServiceOtpDispatcherTest.java
alexbishop1: This offends my grammatical sensibilities too but it’s consistent with the naming of the other tests.

mrlumbu: wouldn't be too much to change other tests too...

PR #1133, opened 2021-10-29
PP-8800 add url to update merchant details

SandorArpa:
## WHAT YOU DID
- Return field in get one service or search services API endpoints
- Accept new field url on PUT /{serviceExternalId}/merchant-details (url is not mandatory)
- Allow patching field url on PATCH /{serviceExternalId}


kbottla (approved):
LGTM!

PR #1243, opened 2022-02-17
PP-9260 Add event subscriber queue message receiver

stephencdaly:
Add a receiver Managed class which starts a ScheduledExecutorService to
poll an SQS queue for event messages and all required config for this.

The event messages expected have the same structure as event messages in
Ledger. The Event model for deserialisation only currently deserialises
the fields that we expect adminusers to consume.


mrlumbu:
Drive by comment: we discussed moving the generic SQS polling logic in to commons - is that still something you're considering?

stephencdaly:
> Drive by comment: we discussed moving the generic SQS polling logic in to commons - is that still something you're considering?

Already moved the core duplicated stuff to commons https://github.com/alphagov/pay-java-commons/commit/764253fffd5d49f85219a86c5a4196dc1630a5c9

There's possibly more that could be generified and shared, but there was a line I had to draw somewhere with how much it would help us to generify and share more.

stephencdaly:
Was going to move some of the config classes to commons too, but was worried this might tie our hands too much if we wanted to change some things for an individual app. Might still be worth doing that..

SandorArpa (approved):
LGTM

PR #1262, opened 2022-03-03
PP-9282 Add ARM64 based Dockerfile for local M1 dev

nlsteers:
## WHAT YOU DID
Added a Dockerfile that will build as a native image on M1 developer Macs, for use by the Pay CLI

### DIFF
--- a/m1/arm64.Dockerfile
+++ b/m1/arm64.Dockerfile
@@ -1,0 +1,32 @@
+FROM eclipse-temurin:11.0.14_9-jre
+
+ARG DNS_TTL=15
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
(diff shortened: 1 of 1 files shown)


m1/arm64.Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM adoptopenjdk/openjdk11:jre-11.0.14_9-ubuntu
```
gidsg, line 1: Could we use the eclipse-temurin image here, to be consistent with what we are doing with Java 17? (this was @heathd's suggestion)

gidsg: Ah wait, maybe it's just that they are keeping version 11 in the old org?

nlsteers: There is an eclipse-temurin 11 image, I think you're right that we should aim to be consistent even if these are dev images only. Have changed to `eclipse-temurin:11.0.14_9-jre`.

stephencdaly (approved):

PR #1271, opened 2022-03-08
PP-9279 add ledger client to adminusers

nlsteers:
## WHAT YOU DID
Added a ledger service to adminusers and updated the application configuration

Co-authored-by: Sandor Arpa <sandor.arpa@digital.cabinet-office.gov.uk>

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/client/ledger/model/LedgerTransaction.java
+++ b/src/main/java/uk/gov/pay/adminusers/client/ledger/model/LedgerTransaction.java
@@ -1,0 +1,37 @@
+package uk.gov.pay.adminusers.client.ledger.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/adminusers/client/ledger/model/LedgerTransaction.java
```diff
@@ -0,0 +1,37 @@
+package uk.gov.pay.adminusers.client.ledger.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
stephencdaly, line 8: Don't think the @JsonInclude annotation is needed - this is just needed if we're serialising this class to a response, which we're not doing.

stephencdaly (approved):
LGTM!

src/main/java/uk/gov/pay/adminusers/client/ledger/model/LedgerTransaction.java
SandorArpa: We shall remove it in the next iteration.

PR #1282, opened 2022-03-14
PP-9379: Add adminusers as provider pact test reusable workflow

jfharden:
## WHAT YOU DID
Add reusable workflow for adminusers-as-provider pact tests. This is the same code as in connector and ledger but for adminusers

## How to test

- See the test workflow in frontend https://github.com/alphagov/pay-frontend/runs/5535387251?check_suite_focus=true (this sets the consumer as selfservice, not frontend, but frontend is the only repo set up so far to be able to execute these and with the pact secrets)

### Logging

N/A

### Documentation

N/A

### DIFF
--- a/.github/workflows/_run-pact-provider-tests.yml
+++ b/.github/workflows/_run-pact-provider-tests.yml
@@ -1,0 +1,61 @@
+name: Adminusers As Provider Pact Tests
+
+on:
+  workflow_call:
+    inputs:
+      consumer:
+        description: Name of the consumer app, e.g. frontend
+        required: true
+        type: string
+      consumer_tag:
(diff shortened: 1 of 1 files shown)


katstevens (approved):

.github/workflows/_run-pact-provider-tests.yml
```diff
@@ -0,0 +1,61 @@
+name: Adminusers As Provider Pact Tests
+
+on:
+  workflow_call:
+    inputs:
+      consumer:
+        description: Name of the consumer app, e.g. frontend
```
oswaldquek, line 7: Wonder if it would be better to use "selfservice" or "products-ui", but technically it shouldn't matter

oswaldquek (approved):

PR #1287, opened 2022-03-15
PP-9236: Run provider contract tests

oswaldquek:
Successful provider contract test build: https://github.com/alphagov/pay-adminusers/runs/5558155471?check_suite_focus=true


.github/workflows/post-merge.yml
```diff
@@ -1,6 +1,7 @@
 name: Post Merge
 
 on:
+  pull_request: #to be delete
```
oswaldquek, line 4: Will delete when PR is approved

.github/workflows/run-tests.yml
```diff
@@ -2,7 +2,7 @@ name: Github Actions Tests
 
 on:
   workflow_call:
-  pull_request:
+#  pull_request:
```
oswaldquek, line 5: Will revert when PR is approved

jfharden (commented):
Great 👍 

I'm happy to approve this. Can you get it into the "final state". (I know post-merge wont run then but the unit and integration tests will which we currently have required and aren't running on this pr. It's demonstrated that the post-merge is good 👍 )

.github/workflows/run-tests.yml
```diff
@@ -29,4 +29,6 @@ jobs:
         run: |
           docker pull govukpay/postgres:11.1
       - name: Run integration tests
-        run: mvn clean verify
+        run: |
+          echo "${{ github.sha }}"
```
jfharden, line 33: Don't think you need this anymore?

.github/workflows/post-merge.yml
```diff
@@ -1,6 +1,7 @@
 name: Post Merge
 
 on:
+  pull_request:
```
jfharden, line 4: This will be removed right?

jfharden (commented):
I approved the other PR this depends on but I think you've left a couple of bits in here

.github/workflows/post-merge.yml
oswaldquek: oh I only deleted the `#to be deleted` comment 😆

oswaldquek: fixed

.github/workflows/run-tests.yml
oswaldquek: Fixed

.github/workflows/post-merge.yml
jfharden: Classic 😄

jfharden (approved):

PR #1291, opened 2022-03-18
[BAU] Speculative fix for Pact Provider Tagging

dj-maisy:
## What?
This is a speculative fix for the Pact Provider tagging issue that we've picked-up. We've discovered that when the workflow is called by other Pact consumers to use Adminusers as the "provider", it re-uses the same Git SHA from the consumer, which is undesired behaviour.

At present, the Checkout Action does not provide an output for this, so we're using a workaround.

## Contributors
Thanks to @jfharden and @katstevens for helping to figure this out.

### DIFF
--- a/.github/workflows/_run-pact-provider-tests.yml
+++ b/.github/workflows/_run-pact-provider-tests.yml
@@ -27,6 +27,9 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
+      - name: Get Provider SHA
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/_run-pact-provider-tests.yml
```diff
@@ -27,6 +27,9 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
+      - name: Get Provider SHA
```
jfharden, line 30: If you do this step after the checkout step the local filesystem will be pay-adminusers so you should be able to just do `git rev-parse HEAD`

dj-maisy: Changed. Let's give it a go.

jfharden (approved):

katstevens (approved):
Let's try it 👍

PR #1294, opened 2022-03-18
PP-9279 Additional logging, model tests and deserialisation bug fixes

nlsteers:
## WHAT YOU DID

- Added additional logging to the event message handler to log information regarding processed notification emails for disputes
- Fixed a bug in DisputeCreatedDetails where a Json property was incorrectly labelled
- Added missing deserialisation test for Dispute Events

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -112,6 +112,10 @@ public class EventMessageHandler {
                     serviceAdmins.stream().map(UserEntity::getEmail).collect(Collectors.toSet()),
                     personalisation
             );
+            logger.info("Processed notification email for disputed transaction",
+                    kv("transaction_id", disputeCreatedEvent.getParentResourceExternalId()),
+                    kv("service_id", service.getId())
+            );
         } else {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -112,6 +112,10 @@ private void handleDisputeCreatedMessage(Event disputeCreatedEvent) throws JsonP
                     serviceAdmins.stream().map(UserEntity::getEmail).collect(Collectors.toSet()),
                     personalisation
             );
+            logger.info("Processed notification email for disputed transaction",
+                    kv("transaction_id", disputeCreatedEvent.getParentResourceExternalId()),
```
stephencdaly, line 116: Could we call this `payment_external_id` to make clear which transaction? It's probably sensible to also log the dispute id too, not sure whether we log this in other places, but might help.

stephencdaly: Possibly we could add some stuff onto the MDC so we have more information on all log lines.

stephencdaly (approved):
Comment about logging, although this can be merged and we can address that in another PR.

PR #1295, opened 2022-03-18
PP-9279 Log exceptions from Notify better

stephencdaly:
Log the exception message when sending an email fails.

Pass along the original exception when throwing a
WebApplicationException so that more information gets to Sentry.


nlsteers (approved):
LGTM

PR #1303, opened 2022-03-22
PP-9278 Fix deserialisation of event_details

stephencdaly:
The `event_details` field in an event message is a String containing JSON, rather than nested JSON in the message. Fix deserialisation to expect this.


SandorArpa (approved):
LGTM

PR #1309, opened 2022-03-23
PP-9281 Fix date format in dispute email

stephencdaly:
Remove the comma to adhere to the GOV.UK style guide
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -37,6 +37,8 @@ import static uk.gov.service.payments.logging.LoggingKeys.SERVICE_EXTERNAL_ID;
 
 public class EventMessageHandler {
 
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern("d MMMM yyyy"); // 9 March 2022
+    
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private final EventSubscriberQueue eventSubscriberQueue;
     private final LedgerService ledgerService;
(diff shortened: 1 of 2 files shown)


kbottla (dismissed):

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -44,7 +44,7 @@
     private final ServiceFinder serviceFinder;
     private final UserServices userServices;
     private final ObjectMapper objectMapper;
-    private final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("dd, LLLL yyyy"); // 09, March 2022
+    private final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("dd LLLL yyyy"); // 09 March 2022
```
alexbishop1, line 47: I think the more familiar `MMMM` would give the same output as `LLLL`. I think we also want `d` rather than `dd` because we do not want a leading zero.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
stephencdaly: thanks, will update

alexbishop1 (approved):

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
stephencdaly: Thanks, I've made both those changes

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -37,14 +37,15 @@
 
 public class EventMessageHandler {
 
+    private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM yyyy"); // 9 March 2022
```
alexbishop1, line 40: Constant should use `SCREAMING_SNAKE_CASE`.

alexbishop1 (dismissed):

kbottla (approved):

PR #1311, opened 2022-03-24
PP-9281 Look up service by gateway account id

stephencdaly:
Due to issues with how ledger is projecting transactions from events,
there is a bug where the service_id on a transaction is overwritten to
null when an event, such as PAYMENT_INCLUDED_IN_PAYOUT doesn't send a
service_id.

This means that for now at least we have to rely on the
gateway_account_id, which is also sent in the event details to look up
the service.


kbottla (approved):
LGTM!

PR #1335, opened 2022-04-06
PP-8827: Endpoint to remove user from service

oswaldquek:
Given that users need to be in the alphagov organisation in GitHub in order to
be authenticated with toolbox, it should be fairly safe to expose an endpoint
to remove a user from a service without requiring a "GovUkPay-User-Context"
header. Note that the original `DELETE
/v1/api/services/{serviceExternalId}/users/{userExternalId}` endpoint is still
valid and is used by selfservice. In order to make it clear `DELETE
/v1/api/toolbox/services/{serviceExternalId}/users/{userExternalId}` is for
toolbox use only we've placed the implementation in
ToolboxEndpointResource.java for clarity.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ToolboxEndpointResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ToolboxEndpointResource.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.adminusers.resources;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.adminusers.service.ServiceServicesFactory;
+
+import javax.inject.Inject;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.Status.NO_CONTENT;
+
+@Path("/v1/api/toolbox")
+public class ToolboxEndpointResource {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ToolboxEndpointResource.class);
+
+    private final ServiceServicesFactory serviceServicesFactory;
+
+    @Inject
+    public ToolboxEndpointResource(ServiceServicesFactory serviceServicesFactory) {
+        this.serviceServicesFactory = serviceServicesFactory;
+    }
+
+    @Path("/services/{serviceExternalId}/users/{userExternalId}")
+    @DELETE
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Response removeUserFromService(@PathParam("serviceExternalId") String serviceExternalId,
+                                          @PathParam("userExternalId") String userExternalId) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ToolboxEndpointResource.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.ws.rs.DELETE;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.Status.NO_CONTENT;
+
+@Path("/v1/api/toolbox")
```
sfount, line 18: Having this explicitly behind a `/toolbox` or `/admin` route feels like a good pattern to avoid other consumers accidentally using this 👍

src/main/java/uk/gov/pay/adminusers/resources/ToolboxEndpointResource.java
```diff
@@ -0,0 +1,41 @@
...
+        this.serviceServicesFactory = serviceServicesFactory;
+    }
+
+    @Path("/services/{serviceExternalId}/users/{userExternalId}")
+    @DELETE
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Response removeUserFromService(@PathParam("serviceExternalId") String serviceExternalId,
+                                          @PathParam("userExternalId") String userExternalId) {
+        LOGGER.info("Toolbox DELETE request - serviceExternalId={}, userExternalId={}", serviceExternalId, userExternalId);
```
sfount, line 36: I think admin users has been set up with structured logging which would allow this to use `kv()` arguments passed in rather than manually templating into the string, I don't think this is a deal breaker but can make parsing the logs much easier

src/main/java/uk/gov/pay/adminusers/resources/ToolboxEndpointResource.java
```diff
@@ -0,0 +1,41 @@
...
+        serviceServicesFactory.serviceUserRemover().remove(userExternalId, serviceExternalId);
```
sfount, line 37: I'm not really familiar with this factory pattern put the call to `serviceUserRemover()` feels like it could be done in the constructor once with the `remove()` method then directly called in the resource -- I'm guessing because of the guice injection this will just be returning an already instantiated singleton so this likely isn't a performance issue. 

This is following the pattern previously set by admin users so this shouldn't be a blocker.

src/main/java/uk/gov/pay/adminusers/service/ServiceUserRemover.java
```diff
@@ -37,6 +37,15 @@ public void remove(String userExternalId, String removerExternalId, String servi
 
         serviceRoleDao.remove(userServiceRoleToRemove);
     }
+    
+    public void remove(String userExternalId, String serviceExternalId) {
```
sfount, line 41: This overloaded pattern is quite neat but as the other remove method is also just a number of strings (3 rather than 2) should this method be named something to show that it's explicitly removing without validation checks? If these functionalities were to be extended I could see accidentally calling the wrong remove by passing in the wrong number of strings. Maybe something like `removeWithoutAdminCheck` ?

sfount (commented):
Code looks great, few questions but nothing blocking

sfount (approved):
Lines up with refined AC linked to Jira ticket

PR #1345, opened 2022-04-11
PP-9456 Remove API specs markdown

kbottla:
## WHAT YOU DID
- Removed API specs markdown as we now have OpenAPI specs for adminusers endpoints


SandorArpa (approved):
LGTM

PR #1370, opened 2022-04-26
PP-9541 Add initial agreements permissions

sfount:
Agreements read and update permissions are currently mapped similarly to
the transactions case worker model. As services start using it if they
need more granular permissions this can be reconsidered.

### DIFF
--- a/src/main/resources/migrations/00079_add_agreement_permission.sql
+++ b/src/main/resources/migrations/00079_add_agreement_permission.sql
@@ -1,0 +1,14 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:agreements_read_update_permission
+INSERT INTO  permissions(id, name, description) VALUES ((SELECT MAX(id) + 1 FROM permissions), 'agreements:read', 'View agreements');
+INSERT INTO  permissions(id, name, description) VALUES ((SELECT MAX(id) + 1 FROM permissions), 'agreements:update', 'Update agreements');
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-and-refund'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-only'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-and-initiate-moto'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00079_add_agreement_permission.sql
```diff
@@ -0,0 +1,13 @@
...
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-and-refund'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-only'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'agreements:update'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'agreements:update'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-and-refund'), (SELECT id FROM permissions WHERE name = 'agreements:update'), NOW(), NOW());
```
alexbishop1, line 13: Do we want view and refund users to be able to modify agreements?

alexbishop1: In that case, I think not granting the permission to view and refund users (leaving it as the preserve of admins) would be best so we do not have to take permissions away from people in the future. Can’t put the genie back in the bottle and all that. 🧞

sfount: Yeah agree @alexbishop1, have pushed to https://github.com/alphagov/pay-adminusers/commit/a512fc4218a4524b83870e6cf69be9e4835c29ec

src/main/resources/migrations/00079_add_agreement_permission.sql
```diff
@@ -0,0 +1,12 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:agreements_read_update_permission
+INSERT INTO  permissions(id, name, description) VALUES ((SELECT MAX(id) + 1 FROM permissions), 'agreements:read', 'View agreements');
+INSERT INTO  permissions(id, name, description) VALUES ((SELECT MAX(id) + 1 FROM permissions), 'agreements:update', 'Update agreements');
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'super-admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'admin'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-and-refund'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
+INSERT INTO  role_permission VALUES ((SELECT id FROM roles WHERE name = 'view-only'), (SELECT id FROM permissions WHERE name = 'agreements:read'), NOW(), NOW());
```
alexbishop1, line 9: Due to uncontrolled proliferation of user roles, we should also give `agreements:read` to the `view-and-initiate-moto` and `view-refund-and-initiate-moto` roles.

alexbishop1 (approved):

PR #1390, opened 2022-05-10
PP-9307 add new pact test

SandorArpa:
## WHAT YOU DID
- add a new pact test with setup for adminusers as consumer and ledger as provider


stephencdaly (approved):
LGTM!

PR #1451, opened 2022-06-16
PP-9576 implement handling of lost dispute event

SandorArpa:
## WHAT YOU DID
- implement the handling of DISPUTE_LOST event type
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -14,12 +14,14 @@ import uk.gov.pay.adminusers.queue.model.Event;
 import uk.gov.pay.adminusers.queue.model.EventMessage;
 import uk.gov.pay.adminusers.queue.model.EventType;
 import uk.gov.pay.adminusers.queue.model.event.DisputeCreatedDetails;
+import uk.gov.pay.adminusers.queue.model.event.DisputeLostDetails;
 import uk.gov.pay.adminusers.service.NotificationService;
 import uk.gov.pay.adminusers.service.ServiceFinder;
 import uk.gov.pay.adminusers.service.UserServices;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -86,6 +90,49 @@ public void processMessages() throws QueueException {
...
+        MDC.put(PAYMENT_EXTERNAL_ID, disputeLostEvent.getParentResourceExternalId());
+        MDC.put(SERVICE_EXTERNAL_ID, disputeLostEvent.getServiceId());
+
+        try {
+            var disputeLostDetails = objectMapper.readValue(disputeLostEvent.getEventDetails(), DisputeLostDetails.class);
+            Service service = serviceFinder.byExternalId(disputeLostEvent.getServiceId())
+                    .orElseThrow(() -> new IllegalArgumentException(format("Service not found [service_id: %s]", disputeLostEvent.getServiceId())));
+            LedgerTransaction transaction = ledgerService.getTransaction(disputeLostEvent.getParentResourceExternalId())
+                    .orElseThrow(() -> new IllegalArgumentException(format("Transaction not found [payment_external_id: %s]", disputeLostEvent.getParentResourceExternalId())));
+            List<UserEntity> serviceAdmins = userServices.getAdminUsersForService(service);
```
stephencdaly, line 104: Feels like there's going to be a lot of duplication here, as we always need to get certain things for all dispute events. 
Could we maybe move the decision of how to handle each dispute event type inside of a shared method called for all dispute notifications that does the adding to the MDC and gets all of the commonly needed things then does something different when it comes to creating the personalisation and sending the email for the different event types?

It's fine if this refactor is done as a next step when adding handling the dispute update events.

stephencdaly (approved):
LGTM! Left a comment about reducing code duplication but refactor could be done in the next PR if appropriate

PR #1453, opened 2022-06-17
PP-9576 refactor dispute lost event handling

SandorArpa:
## WHAT YOU DID
- refactor DISPUTE_LOST event handling to use gateway account id. This is needed because disputes can be opened up to 120 days and sending service id was only fixed on 24/03
- also refactor a few code repetitions
- refactor tests as well


stephencdaly (approved):
LGTM!

PR #1506, opened 2022-07-20
PP-9677 Remove fields from dispute lost email personalisation

stephencdaly:
Remove the disputeFee and disputeAmount fields from the personalisation
as these are not yet used by the Notify template.

The `fee` and `net_amount` fields will only be sent in the DISPUTE_LOST
event from connector when we are charging for disputes.

The story to update the template with information about the amount
charged for lost disputes can add these back in where appropriate.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -161,9 +161,6 @@ public class EventMessageHandler {
                 Map<String, String> personalisation = getMinimumRequiredPersonalisation(disputeLostDetails.getGatewayAccountId(),
                         disputeLostEvent.getParentResourceExternalId());
 
-                personalisation.put("disputedAmount", convertPenceToPounds.apply(disputeLostDetails.getAmount()).toString());
-                personalisation.put("disputeFee", convertPenceToPounds.apply(disputeLostDetails.getFee()).toString());
-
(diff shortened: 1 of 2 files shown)


iqbalgds (approved):

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -161,9 +161,6 @@ private void handleDisputeLostMessage(Event disputeLostEvent) throws JsonProcess
                 Map<String, String> personalisation = getMinimumRequiredPersonalisation(disputeLostDetails.getGatewayAccountId(),
                         disputeLostEvent.getParentResourceExternalId());
 
-                personalisation.put("disputedAmount", convertPenceToPounds.apply(disputeLostDetails.getAmount()).toString());
```
kbottla, line 164: Do we really want to remove these? We will need these in a few weeks time (for PP-9841).

stephencdaly: Either we remove, or we make sure we only include the fee when it's there otherwise it will break. I opted to remove as it makes sense to add that logic in PP-9841. Not difficult at all to add back in.

stephencdaly: Guess I could just add in that logic now, as the template is largely defined

kbottla (approved):

PR #1516, opened 2022-07-26
PP-9841: Use different Notify template when a lost dispute has a fee

alan-gds:
The NotificationService chooses between two different templates based on whether the personalisation info includes details of a fee.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -158,7 +158,7 @@ public class EventMessageHandler {
             MDC.put(GATEWAY_ACCOUNT_ID, disputeLostDetails.getGatewayAccountId());
 
             if (shallSendDisputeUpdatedEmail(disputeLostEvent)) {
-                Map<String, String> personalisation = getMinimumRequiredPersonalisation(disputeLostDetails.getGatewayAccountId(),
+                Map<String, String> personalisation = getPersonalisationForDisputeLost(disputeLostDetails,
                         disputeLostEvent.getParentResourceExternalId());
 
                 sendEmailNotificationToServiceAdmins(disputeLostEvent.getEventType(), disputeLostDetails.getGatewayAccountId(),
@@ -180,6 +180,17 @@ public class EventMessageHandler {
                         service.getMerchantDetails().getName() : service.getName(),
                 "serviceName", service.getName(),
                 "serviceReference", transaction.getReference()));
+    }
+
+    private Map<String, String> getPersonalisationForDisputeLost(DisputeLostDetails details, String parentResourceExternalId) {
+        var personalisation = getMinimumRequiredPersonalisation(details.getGatewayAccountId(), parentResourceExternalId);
+
+        if (details.getFee() != null) {
+            personalisation.put("disputedAmount", convertPenceToPounds.apply(details.getAmount()).toString());
+            personalisation.put("disputeFee", convertPenceToPounds.apply(details.getFee()).toString());
+        }
+
+        return personalisation;
     }
 
     private boolean shallSendDisputeUpdatedEmail(Event disputeLostEvent) {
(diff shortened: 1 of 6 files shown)


alan-gds:
Integration tests are failing, most likely because I've yet not deployed the infra changes I made earlier...

src/test/resources/config/test-it-config.yaml
```diff
@@ -73,6 +73,7 @@ notify:
   liveAccountCreatedEmailTemplateId: ${NOTIFY_LIVE_ACCOUNT_CREATED_EMAIL_TEMPLATE_ID:-pay-notify-live-account-created-email-template-id}
   stripeDisputeCreatedEmailTemplateId: ${NOTIFY_STRIPE_DISPUTE_CREATED_EMAIL_TEMPLATE_ID:-pay-notify-stripe-dispute-created-email-template-id}
   stripeDisputeLostEmailTemplateId: ${NOTIFY_STRIPE_DISPUTE_LOST_EMAIL_TEMPLATE_ID:-pay-notify-stripe-dispute-lost-email-template-id}
+stripeDisputeLostAndServiceChargedEmailTemplateId: ${NOTIFY_STRIPE_DISPUTE_LOST_AND_SERVICE_CHARGED_EMAIL_TEMPLATE_ID:-pay-notify-stripe-dispute-lost-email-template-id}
```
kbottla, line 76: Integration tests could be failing due to indentation here. 
```suggestion
  stripeDisputeLostAndServiceChargedEmailTemplateId: ${NOTIFY_STRIPE_DISPUTE_LOST_AND_SERVICE_CHARGED_EMAIL_TEMPLATE_ID:-pay-notify-stripe-dispute-lost-email-template-id}
```

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -182,6 +182,17 @@ private Map<String, String> getMinimumRequiredPersonalisation(String gatewayAcco
                 "serviceReference", transaction.getReference()));
     }
 
+    private Map<String, String> getMinimumPersonalisationForDisputeLost(DisputeLostDetails details, String parentResourceExternalId) {
```
kbottla, line 185: Any reason separate method is needed ? Or maybe rename method to `getPersonalisationForDisputeLost` .
`getMinimumPersonalisationForDisputeLost` indicates the result will be personalised further which is not the case.

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -182,6 +182,17 @@ private Map<String, String> getMinimumRequiredPersonalisation(String gatewayAcco
...
+        var personalisation = getMinimumRequiredPersonalisation(details.getGatewayAccountId(), parentResourceExternalId);
+
+        if (details.getFee() != 0) {
```
kbottla, line 188:
```suggestion
        if (details.getFee() !=null) {
```

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -182,6 +182,17 @@ private Map<String, String> getMinimumRequiredPersonalisation(String gatewayAcco
...
+            personalisation.put("disputedAmount", details.getAmount().toString());
```
kbottla, line 189:
```suggestion
            personalisation.put("disputedAmount", convertPenceToPounds.apply(details.getAmount()).toString());
```

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -182,6 +182,17 @@ private Map<String, String> getMinimumRequiredPersonalisation(String gatewayAcco
...
+            personalisation.put("disputeFee", details.getFee().toString());
```
kbottla, line 190:
```suggestion
            personalisation.put("disputeFee", convertPenceToPounds.apply(details.getFee()).toString());
```

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
alan-gds: If we only had one method, then calls for every type of dispute would have to supply a DisputeLostDetails argument.  Happy to rename the second method as suggested though.

src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
alan-gds: Oops.  I failed to spot that this was a 'Long', not a 'long'.

alan-gds:
Thanks for the review.  I've made the suggested changes, aside from removing the new personalisation method (as I think it's better not to pollute the signature of the other method).

kbottla (approved):

PR #1521, opened 2022-07-28
PP-9671 Handle unknown event types

kbottla:
## WHAT YOU DID
- Handle unknown event types better. The current way of getting EventType enum throws an exception if an event type is not in enum constants

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
@@ -71,7 +71,7 @@ public class EventMessageHandler {
         List<EventMessage> eventMessages = eventSubscriberQueue.retrieveEvents();
         for (EventMessage message : eventMessages) {
             try {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/adminusers/queue/event/EventMessageHandler.java
```diff
@@ -90,7 +90,7 @@ public void processMessages() throws QueueException {
                         handleDisputeWonMessage(message.getEvent());
                         break;
                     default:
-                        logger.warn("Unknown event type: {}", eventType);
+                        logger.warn("Unknown event type: {}", message.getEvent().getEventType());
```
stephencdaly, line 93: Should we stop logging at warn level here? DISPUTE_INCLUDED_IN_PAYOUT events will reach adminusers but adminusers will never care about them.

kbottla: yeah, don't need to be at the warning level. Updated to `info`.

stephencdaly (approved):

PR #1546, opened 2022-08-11
PP-9292 Run consumer tests as part of integration test suite

stephencdaly:
Currently consumer contract tests for SQS event messages are run as part of a separate suite. This is run as a separate concourse job.

In reality, these are integration tests that also generate a Pact file to be verified against the provider. We run consumer tests that are unit tests as part of the unit test suite, so it makes sense to run these as part of the integration test suite.

Moving these to be run as part of the integration test suite will simplify our concourse pipeline as then only one job is responsible for publishing pact files.

An additional benefit is individual tests no longer need to be added to a suite file, which can be missed.

Rename *EventQueueContactTest -> *EventQueueConsumerIT to achieve this.


SandorArpa (approved):
LOKTM

PR #1595, opened 2022-09-15
PP-10041: Validate OTP during completion step

alan-gds:
## WHAT YOU DID
* Validate the OTP when the `/{code}/complete` endpoint is invoked, so long as the request is not for an 'existing user' journey.
* At present this will be repeating the validation of the OTP, which will already have been performed by an earlier call, but we will remove the earlier duplicate soon once we have unified the divergent journey flows.
* The `Invite` model now represents its `type` field as an instance of the `InviteType` enum, rather than as a string.
  * I'm not sure whether this is legit.  I checked where the field gets used, but I am wondering whether the class is ever used in anything reflectiony; there are annotations on a lot of the fields but I'm not sure what they relate to - plausibly json serialisation/deserialisation, in which case this change might cause an issue?
* There is an opportunity to refactor the part of `InviteResource#completeInvite` which deals with actually completing the invite (the lengthy `return` statement beginning on line 136).  I think I will leave that until we have added a third value to the `inviteType` enum though.

## How to test

- Make sure the `completeInvite` method still succeeds/fails correctly for the relevant journeys (journeys 1 and 2 as described in the task).
- Consider whether the change to the `Invite` model is safe.

### Logging

### Documentation

### DIFF
--- a/openapi/adminusers_spec.yaml
+++ b/openapi/adminusers_spec.yaml
@@ -1197,6 +1197,9 @@ components:
           example: +440787654534
         type:
           type: string
+          enum:
+          - USER
+          - SERVICE
           example: service
         user_exist:
           type: boolean
--- a/src/main/java/uk/gov/pay/adminusers/model/Invite.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Invite.java
@@ -23,7 +23,7 @@ public class Invite {
     private Integer attemptCounter = 0;
 
     private List<Link> links = new ArrayList<>();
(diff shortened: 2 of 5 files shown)


alan-gds:
Change of plan.  We are no longer going to integrate OTP validation into the completion step; it adds too much complexity to defining the test coverage.  Instead, we are going to keep them separate, and make the other journey (that currently performs them together) perform them separately too.  I will close this PR and revert the previous one that was done for this task.

PR #1639, opened 2022-10-05
PP-10041: Add new InviteType constants and refactor invite completers

alan-gds:
## WHAT YOU DID
* Add new InviteType constants, which will replace the old ones (the new ones have better names and cover the full range of journeys).  This PR adds the new constants but does not start using them.
* Rename the InviteCompleter implementations to reduce ambiguity, and add new InviteCompleter to handle the 'new user, existing service' journey.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
@@ -23,7 +23,7 @@ public class InviteRouter {
         return routeIfExist(inviteCode,
                 inviteEntity -> {
                     boolean isServiceType = inviteEntity.isServiceType();
-                    InviteCompleter inviteCompleter = isServiceType ? inviteServiceFactory.completeServiceInvite() : inviteServiceFactory.completeExistingUserInvite();
+                    InviteCompleter inviteCompleter = isServiceType ? inviteServiceFactory.completeSelfSignupInvite() : inviteServiceFactory.completeExistingUserInvite();
                     return Optional.of(Pair.of(inviteCompleter, isServiceType));
                 });
     }
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
@@ -23,7 +23,7 @@ public class InviteRouter {
         return routeIfExist(inviteCode,
                 inviteEntity -> {
                     boolean isServiceType = inviteEntity.isServiceType();
-                    InviteCompleter inviteCompleter = isServiceType ? inviteServiceFactory.completeServiceInvite() : inviteServiceFactory.completeUserInvite();
+                    InviteCompleter inviteCompleter = isServiceType ? inviteServiceFactory.completeServiceInvite() : inviteServiceFactory.completeExistingUserInvite();
                     return Optional.of(Pair.of(inviteCompleter, isServiceType));
                 });
     }
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
@@ -19,12 +19,12 @@ public class InviteRouter {
         this.inviteDao = inviteDao;
     }
 
-    public Optional<Pair<InviteCompleter, Boolean>> routeComplete(String inviteCode) {
+    public Optional<InviteCompleter> routeComplete(String inviteCode) {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
```diff
@@ -19,13 +20,23 @@ public InviteRouter(InviteServiceFactory inviteServiceFactory, InviteDao inviteD
...
+                        case USER:
+                        case EXISTING_USER_INVITED_TO_EXISTING_SERVICE:
+                            return Optional.of(Pair.of(inviteServiceFactory.completeExistingUserInvite(), false));
+                        case NEW_USER_INVITED_TO_EXISTING_SERVICE:
+                            return Optional.of(Pair.of(inviteServiceFactory.completeNewUserExistingServiceInvite(), true));
+                        default:
+                            throw new WebApplicationException(String.format("Unrecognised invite type: %s", inviteEntity.getType()));
+                    }
+                })
+                .map(Pair::getLeft);
```
stephencdaly, line 39: If we're calling `getLeft` does the switch statement need to return a Pair at all? Are we using the right value?

src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
```diff
@@ -0,0 +1,77 @@
...
+                            .map(serviceRole -> serviceRole.getService().getExternalId())
+                            .collect(Collectors.joining(", "));
+
+                    LOGGER.info("User created successfully from invitation [{}] for services [{}]", userEntity.getExternalId(), serviceIds);
+
+                    Invite invite = linksBuilder.addUserLink(userEntity.toUser(), inviteEntity.toInvite());
+                    InviteCompleteResponse response = new InviteCompleteResponse(invite);
+                    response.setUserExternalId(userEntity.getExternalId());
+
+                    return Optional.of(response);
```
stephencdaly, line 72: Looks like we don't need to return an Optional, as it will always have a value. If the other completers don't need to return an Optional, we could refactor this in a following PR?

stephencdaly (dismissed):

src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
alan-gds: You're right; I've now rationalised this and squashed it into the existing rationalisation commit.

src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
alan-gds: One of the other completers does return an Optional, but I think it makes more sense for it to work that same way that my new completer does.  I'll refactor it as part of this PR, so that we don't have inconsistent code in the interim.

alan-gds: On second thoughts, it means changing the interface...I'll go along with your suggestion of using a separate PR. :)

src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
```diff
@@ -0,0 +1,77 @@
...
+                    UserEntity userEntity = inviteEntity.mapToUserEntity();
+                    userDao.persist(userEntity);
+                    inviteEntity.setDisabled(Boolean.TRUE);
+                    inviteDao.merge(inviteEntity);
+                    
+                    String serviceIds = userEntity.getServicesRoles().stream()
+                            .map(serviceRole -> serviceRole.getService().getExternalId())
+                            .collect(Collectors.joining(", "));
+
+                    LOGGER.info("User created successfully from invitation [{}] for services [{}]", userEntity.getExternalId(), serviceIds);
```
sfount, line 66: Does admin users do any `kv()` structured logging? as this is a new thing I wonder if it would be good to start leaning on that instead of fixed strings

sfount (approved):
Code looks good to me, am I right in thinking that `NEW_USER_INVITED_TO_EXISTING_SERVICE` won't currently be used until the endpoints that the admin tool call are changed? Got a few things we can chat about but otherwise this looks well thought through

alan-gds:
Thanks for the feedback; I will be raising a further refactoring PR soon, and will deal with these suggestions in that.

PR #1640, opened 2022-10-05
PP-10041: Follow-up from #1639

alan-gds:
## WHAT YOU DID
* Don't return Optionals from InviteCompleter
* Don't use WebApplicationException in InviteRouter
* Use structured logging in NewUserExistingServiceInviteCompleter

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
@@ -117,9 +117,8 @@ public class InviteResource {
 
         return inviteServiceFactory.inviteCompleteRouter().routeComplete(inviteCode)
                 .map(inviteCompleter -> {
-                    return inviteCompleter.withData(inviteCompleteRequestFrom(payload)).complete(inviteCode)
-                            .map(inviteCompleteResponse -> Response.status(OK).entity(inviteCompleteResponse).build())
-                            .orElseGet(() -> Response.status(NOT_FOUND).build());
+                    InviteCompleteResponse inviteCompleteResponse = inviteCompleter.withData(inviteCompleteRequestFrom(payload)).complete(inviteCode);
+                    return Response.status(OK).entity(inviteCompleteResponse).build();
                 })
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
--- a/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
@@ -106,21 +106,18 @@ public class InviteResource {
                     @ApiResponse(responseCode = "404", description = "Not found")
             }
     )
-    public Response completeInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode,
+    public InviteCompleteResponse completeInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode,
                                    @Parameter(schema = @Schema(implementation = InviteCompleteRequest.class))
                                            JsonNode payload) {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -117,9 +117,8 @@ public Response completeInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0r
 
         return inviteServiceFactory.inviteCompleteRouter().routeComplete(inviteCode)
                 .map(inviteCompleter -> {
-                    return inviteCompleter.withData(inviteCompleteRequestFrom(payload)).complete(inviteCode)
-                            .map(inviteCompleteResponse -> Response.status(OK).entity(inviteCompleteResponse).build())
-                            .orElseGet(() -> Response.status(NOT_FOUND).build());
+                    InviteCompleteResponse inviteCompleteResponse = inviteCompleter.withData(inviteCompleteRequestFrom(payload)).complete(inviteCode);
+                    return Response.status(OK).entity(inviteCompleteResponse).build();
```
sfount, line 121: Bonus cleanup -- DropWizard allows you to set the entity that should be sent from a successful call to the resource as the method return type, in my view this makes the code cleaner and I think it means the OpenAPI spec has more to work with etc.

This might look something like: 

```java
public InviteCompleteResponse completeInvite() {
  // ...
  return inviteServiceFactory.inviteCompleteRouter().routeComplete(inviteCode)
    .map(inviteCompleter -> inviteCompleter.withData(inviteCompleteRequestFrom(payload)).complete(inviteCode))
    .orElseThrow(() -> new WebApplicationException(Response.Status.NOT_FOUND))
}
```

src/main/java/uk/gov/pay/adminusers/service/ExistingUserInviteCompleter.java
```diff
@@ -46,14 +45,14 @@ public Optional<InviteCompleteResponse> complete(String inviteCode) {
...
                                     response.setServiceExternalId(inviteEntity.getService().getExternalId());
-                                    return Optional.of(response);
+                                    return response;
                                 } else {
                                     throw internalServerError(format("Attempting to complete user subscription to a service for a non existent service. invite-code = %s", inviteEntity.getCode()));
                                 }
-                            }).orElseGet(() -> {
+                            }).orElseThrow(() -> {
                                 throw internalServerError(format("Attempting to complete user subscription to a service for a non existent user. invite-code = %s", inviteEntity.getCode()));
                             });
```
sfount, line 54: `orElseThrow` with a throw inside seems a little confusing to me -- that might just be me! 

Does this work?

```suggestion
                            }).orElseThrow(() -> internalServerError(format("Attempting to complete user subscription to a service for a non existent user. invite-code = %s", inviteEntity.getCode()));
```

alan-gds: No, this is definitely me, failing to Java!  Will fix.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -106,22 +106,18 @@ public Response getInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @
                     @ApiResponse(responseCode = "404", description = "Not found")
             }
     )
-    public Response completeInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode,
+    public InviteCompleteResponse completeInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode,
```
sfount, line 109: 🍬

sfount (approved):
Code reads a lot clearer, really good stuff!

PR #1643, opened 2022-10-06
PP-10041: Make InviteType parsing more robust

alan-gds:
Really this should have been done as part of #1639, but I overlooked it.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/InviteType.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/InviteType.java
@@ -1,35 +1,21 @@
 package uk.gov.pay.adminusers.model;
 
-import static java.lang.String.format;
-
 public enum InviteType {
 
     // Old values; these are still in use.
-    USER("user"),
-    SERVICE("service"),
+    USER,
+    SERVICE,
 
     // New values; these are not used yet.
-    EXISTING_USER_INVITED_TO_EXISTING_SERVICE("existing_user_invited_to_existing_service"),
-    NEW_USER_INVITED_TO_EXISTING_SERVICE("new_user_invited_to_existing_service"),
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/adminusers/model/InviteType.java
```diff
@@ -23,13 +26,11 @@ public String getType() {
...
-            return SERVICE;
-        } else {
-            throw new RuntimeException(format("invalid invite type: [%s]", inviteType));
-        }
+    public static InviteType from(String typeString) {
+        Optional<InviteType> match = Arrays.stream(InviteType.values())
+                .filter(inviteType -> inviteType.type.equals(typeString))
+                .findAny();
+        
+        return match.orElseThrow(() -> new RuntimeException(format("invalid invite type: [%s]", typeString)));
```
alexbishop1, line 34: You can avoid the `match` local variable:

```java
return Arrays.stream(InviteType.values())
                .filter(inviteType -> inviteType.type.equals(typeString))
                .findAny()
                .orElseThrow(() -> new RuntimeException(format("invalid invite type: [%s]", typeString)));
```
```

alexbishop1 (dismissed):

alexbishop1:
Yeah, that’s good (assuming we don’t expect unknown values). I don’t know where `type` is coming from but if you make that always be upper-case then things would be a bit simpler.

alan-gds:
> Yeah, that’s good (assuming we don’t expect unknown values). I don’t know where `type` is coming from but if you make that always be upper-case then things would be a bit simpler.

What used to be `type` (and is now just the `typeString` parameter) corresponds to the values that are written to the database - we can't change these.  I'll be raising another PR later that uses Jackson to do all of this, but that's a more radical change, so Steve F suggested we get this one merged first.

alexbishop1 (approved):

PR #1644, opened 2022-10-06
PP-10041: Use Jackson serialisation for InviteType

alan-gds:
The custom serialisation/deserialisation code for InviteType is now implemented in InviteTypeConverter, rather than inside the InviteType enum itself.  This is more idiomatic.  Also, it means we can now exclusively use the enum type in the rest of the codebase, instead of messing around with its string representation.  Jackson/JPA will deal with serialisation and deserialisation as needed; that's their job.

Assuming that I'm right about how Jackson/JPA will behave, this PR should be functionally equivalent to the recently merged #1643.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteTypeConverter.java
+++ b/src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteTypeConverter.java
@@ -10,12 +10,12 @@ public class InviteTypeConverter implements AttributeConverter<InviteType, Strin
 
     @Override
     public String convertToDatabaseColumn(InviteType inviteType) {
-        return inviteType.getType();
+        return inviteType.name().toLowerCase();
     }
 
     @Override
     public InviteType convertToEntityAttribute(String string) {
-        return InviteType.from(string);
+        return InviteType.valueOf(string.toUpperCase());
     }
 
 }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteTypeConverter.java
```diff
@@ -10,12 +10,12 @@ public class InviteTypeConverter implements AttributeConverter<InviteType, Strin
 
     @Override
     public String convertToDatabaseColumn(InviteType inviteType) {
-        return inviteType.getType();
+        return inviteType.name().toLowerCase();
```
oswaldquek, line 13: Probably from my lack of Java devving these days but why is it to lower case here?

oswaldquek: What's JSP? Surely not Java/Jakarta Server Pages 😂

oswaldquek (approved):

src/main/java/uk/gov/pay/adminusers/persistence/entity/InviteTypeConverter.java
alan-gds: Nope, it's me using the wrong acronym when I mean JPA. 😆  I'll edit the various comments where I've gotten that wrong...

alan-gds:
This isn't going to work out.  We perform serialisation for two different purposes - for persistence in the database, and for passing information in REST requests - and they're both currently tied to the same implementation.  It's not worth the effort to unpick them, and we'd still need to use a custom converter for the JPA serialisation anyway to keep it compatible with what's already in the DB.  I'm closing the PR.

PR #1648, opened 2022-10-10
PP-10051 Generate new OTP secrets in a consistent way

stephencdaly:
- Remove parsing of an `otp_key` field from the create invite requests. This is never provided by consumers of the API.

- Previously, we were generating a random UUID for use as the otp_key on created invites/users. This is not compatible with the secrets we generate for use by authenticator apps, which uses the GoogleAuthenticator library to generate a base 32 encoded secret that can be used by authenticator apps.

- Make it so that whenever a User/Invite is created, the GoogleAuthenticator library is used to generate the secret, so that all OTP secrets are generated in a consistent way.

- Currently, when generating OTP codes, we check which type of secret the User/Invite has, and generate the code using a different method accordingly. OTP secrets generated using the old method will remain in the database, so for now we have to keep both methods of generating codes. Once all secrets have been updated in the database to be base 32 encoded, we will then be able to remove the old way of generating OTP codes.


alan-gds (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #1655, opened 2022-10-11
PP-10041: Prepare to switch to new InviteType constants

alan-gds:
Before we can switch over to using the new values, we need to update the places where they are used:
* Remove isServiceType() and isUserType() methods from InviteEntity
* Temporarily replace the above with isSelfService() and isExistingUserExistingService methods on InviteType - we will get rid of these once the old enum values no longer need to be handled
* Use the enum values (old and new) to choose which InviteOtpDispatcher to use in InviteRouter
  * Also, refactor InviteRouter a bit to get rid of a pointless Optional type
* Map the new enum values to Notify templates for sending OTP SMS messages 
* Support use of the new enum values in InviteService
* Add test coverage for the new enum values

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
@@ -2,6 +2,7 @@ package uk.gov.pay.adminusers.service;
 
 import com.google.inject.Inject;
 import org.apache.commons.lang3.tuple.Pair;
+import uk.gov.pay.adminusers.model.InviteType;
 import uk.gov.pay.adminusers.persistence.dao.InviteDao;
 import uk.gov.pay.adminusers.persistence.entity.InviteEntity;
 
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/adminusers/service/InviteRouter.java
```diff
@@ -25,30 +26,42 @@ public Optional<InviteCompleter> routeComplete(String inviteCode) {
...
-                            return Optional.of(inviteServiceFactory.completeNewUserExistingServiceInvite());
+                            return inviteServiceFactory.completeNewUserExistingServiceInvite();
                         default:
                             throw new IllegalArgumentException(String.format("Unrecognised invite type: %s", inviteEntity.getType()));
                     }
                 });
     }
 
+    /**
+     * @return an optional pair consisting of: the InviteOtpDispatcher to use, and a boolean flag to indicate whether the OTP requires validation during invite completion.
```
stephencdaly, line 42: I think the boolean flag is actually returning whether or not the invite type is user or service, and then using that to determine what it should validate on the request payload.

The code for that is horrible and we can do some work later down the line to tidy that up. However for now, would it make things a tiny bit clearer to return a Pair<InviteOtpDispatcher, InviteType> - it should make the refactor a bit easier later on.

stephencdaly (commented):
Just a small comment

stephencdaly (approved):
This PR can be merged and we can do a refactor to get rid of the Pairs in the next PR

PR #1657, opened 2022-10-12
PP-10041 refactor getting invite

stephencdaly:
PP-10041 Refactor InviteRouter#routeOtpDispatch

Refactor the InviteRouter#routeOtpDispatch to accept an InviteEntity
rather than a code. Instead, look up the InviteEntity in the resource.
This simplifies the code as InviteRouter#routeOtpDispatch no longer
needs to return a Pair with a boolean indicating whether or not the
telephone number and password should be expected in the HTTP request
body.

PP-10041 Refactor invite completer

Refactor invite completer resource so that the resource gets the
InviteEntity once and passes it to the code that needs to use it. This
avoids having to look up the invite multiple times and also simplifies
return types.

with @alan-gds

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -357,7 +357,7 @@
         "filename": "src/test/java/uk/gov/pay/adminusers/resources/InviteResourceGenerateOtpIT.java",
         "hashed_secret": "833b5c8e76b19b617849d9a67a5980f08c784882",
         "is_verified": false,
-        "line_number": 24
+        "line_number": 26
       }
     ],
     "src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java": [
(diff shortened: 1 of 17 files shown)


alan-gds (commented):
Should we add test cases to the integration tests for InviteResource to cover the cases where an invalid invite code is provided?  This used to be covered by the unit tests for the completers/otp dispatchers but that logic is now handled by the resource, so we've lost some coverage.

alan-gds (approved):
Stephen has spotted that we do in fact have coverage for non-existent invite codes, in https://github.com/alphagov/pay-adminusers/blob/master/src/test/java/uk/gov/pay/adminusers/resources/InviteResourceUserCompleteIT.java#L63 and https://github.com/alphagov/pay-adminusers/blob/master/src/test/java/uk/gov/pay/adminusers/resources/InviteResourceGenerateOtpIT.java#L44, which resolves my concern in my previous comment.  So, these changes all look good to me. 👍

PR #1660, opened 2022-10-13
PP-10048 Refactor ExistingUserOtpDispatcher

stephencdaly:
Simplify the code in ExistingUserOtpDispatcher by accepting the UserEntity as a parameter and looking up the UserEntity at the resource level.

If sending a 2FA code fails because an OTP key isn't present on the user, throw a WebApplication rather than returning an empty Optional. This also allows for a more specific error message to be returned in the HTTP response.

These changes reduce code nesting and avoid returning Optionals.


alan-gds (approved):
LGTM 👍

PR #1661, opened 2022-10-13
PP-10048 Prepare for telephone number being nullable

stephencdaly:
The `telephone_number` column on the `users` table is going to be made nullable. Prepare for this by returning an Optional from the getter on the UserEntity, and handling it not being present by throwing an exception that is mapped to a 412 error response.


alan-gds (approved):
LGTM

PR #1663, opened 2022-10-13
PP-10048 Return error for reset second factor

stephencdaly:
Return a 412 error when the /v1/api/users/{userExternalId}/reset-second-factor endpoint is called if the user does not have a telephone number set.

This is because We don't want to allow resetting a user's method from APP to SMS in Toolbox if the user does not have a telephone number set.


alan-gds (approved):
LGTM

PR #1670, opened 2022-10-17
PP-10041: Start using new InviteType enum values

alan-gds:
## WHAT YOU DID
Start using the new InviteType enum values.  The old values will still be present in the database on older invites, until they either expire after 2 days or we do a migration - so, we're not removing support for the old values yet.


src/main/java/uk/gov/pay/adminusers/service/UserInviteCreator.java
```diff
@@ -92,6 +94,7 @@ public Optional<Invite> doInvite(InviteUserRequest inviteUserRequest) {
         }
 
         ServiceEntity serviceEntity = serviceEntityOptional.get();
+        InviteType inviteType = existingUser.isPresent() ? EXISTING_USER_INVITED_TO_EXISTING_SERVICE : NEW_USER_INVITED_TO_EXISTING_SERVICE;
```
stephencdaly, line 97: Would be good to have a unit test covering this

stephencdaly (approved):
LGTM!

PR #1671, opened 2022-10-17
PP-10041: Add v2 /otp/validate endpoint to InviteResource

alan-gds:
This is a copy of the existing v1 /otp/validate/service endpoint, which will be removed later once we have switched to using this new one.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
@@ -50,11 +50,9 @@ import static javax.ws.rs.core.Response.Status.OK;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.internalServerError;
 
-@Path(InviteResource.INVITES_RESOURCE)
+@Path("/")
 @Tag(name = "Invites")
 public class InviteResource {
-
-    public static final String INVITES_RESOURCE = "/v1/api/invites";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(InviteResource.class);
     private static final int MAX_LENGTH_CODE = 255;
@@ -71,7 +69,7 @@ public class InviteResource {
     }
 
     @GET
-    @Path("/{code}")
+    @Path("/v1/api/invites/{code}")
     @Produces(APPLICATION_JSON)
     @Operation(
             summary = "Find invite for invite code",
@@ -94,7 +92,7 @@ public class InviteResource {
     }
 
     @POST
-    @Path("/{code}/complete")
+    @Path("/v1/api/invites/{code}/complete")
     @Produces(APPLICATION_JSON)
     @Operation(
             summary = "Completes the invite by creating user/service and invalidating the invite code",
@@ -135,7 +133,7 @@ public class InviteResource {
     }
 
     @POST
-    @Path("{code}/otp/generate")
+    @Path("/v1/api/invites/{code}/otp/generate")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     @Operation(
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -71,7 +72,7 @@ public InviteResource(InviteService service, InviteRequestValidator inviteValida
     }
 
     @GET
-    @Path("/{code}")
+    @Path(INVITES_RESOURCE_V1 + "/{code}")
```
stephencdaly, line 75: I think these days we prefer to just inline the path strings rather than using constants, as it makes it easier to read and search.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -324,6 +326,29 @@ public Response validateOtpKeyForService(@Parameter(schema = @Schema(implementat
                         .orElseGet(() -> Response.status(OK).build()));
     }
 
+    @POST
+    @Path(INVITES_RESOURCE_V2 + "/otp/validate")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Validates OTP for the invite - part of creating service",
```
stephencdaly, line 334: Can remove "- part of creating service" from this summary

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java
```diff
@@ -270,4 +270,46 @@ public void validateServiceOtpKey_shouldFailWith401_whenInvalidOtp() throws Exce
...
+                .when()
+                .body(mapper.writeValueAsString(sendRequest))
+                .contentType(JSON)
+                .post(V2_INVITES_VALIDATE_OTP_RESOURCE_URL)
+                .then()
+                .statusCode(OK.getStatusCode());
+    }
+
+    @Test
+    public void validateOtpKey_shouldFailWith401_whenInvalidOtp() throws Exception {
```
stephencdaly, line 295: It looks like there wasn't an equivalent test for `/v1/otp/validate/service` like validateOtp_shouldFailAndLockInvite_whenInvalidOtpAuthCode_ifMaxRetryExceeded`.

Is this adequately covered by unit tests, or is it worth adding an equivalent test for the v2 endpoint?

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
alan-gds: Good spot!

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
alan-gds: 👍

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java
alan-gds: I think it makes sense to add an integration test.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -50,11 +50,12 @@
...
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.internalServerError;
 
-@Path(InviteResource.INVITES_RESOURCE)
+@Path("/")
 @Tag(name = "Invites")
 public class InviteResource {
 
-    public static final String INVITES_RESOURCE = "/v1/api/invites";
+    public static final String INVITES_RESOURCE_V1 = "/v1/api/invites";
+    public static final String INVITES_RESOURCE_V2 = "/v2/api/invites";
```
stephencdaly, line 58: Can we remove these constants and in-line all the usages. This is something we've gradually done in other apps, so would be good to make a start here.

stephencdaly (approved):
LGTM!

PR #1700, opened 2022-10-28
PP-10260 Refactor complete invite resource

stephencdaly:
Accept a POJO in the method signature for deserialising the request body.

Refactor InviteCompleter to remove the `withData` method and instead pass in the CompleteInviteRequest as an argument of `InviteCompleter#complete` to make it easier to understand. The `CompleteInviteRequest` will be null if the request was made with no body.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/CompleteInviteRequest.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/CompleteInviteRequest.java
@@ -5,19 +5,21 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Schema;
 
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
-public class InviteCompleteRequest {
+public class CompleteInviteRequest {
 
-    public static final String FIELD_GATEWAY_ACCOUNT_IDS = "gateway_account_ids";
-
-    @ArraySchema(schema = @Schema(implementation = String.class, example = "1,2",
+    @ArraySchema(schema = @Schema(example = "1",
             description = "gateway_account_ids that needs to be associated for the new service. Only applicable for invite type `service`"))
-    private List<String> gatewayAccountIds = new ArrayList<>();
-
-    public void setGatewayAccountIds(List<String> gatewayAccountIds) {
+    private List<String> gatewayAccountIds = Collections.emptyList();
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/adminusers/service/SelfSignupInviteCompleter.java
```diff
@@ -52,8 +53,8 @@ public InviteCompleteResponse complete(InviteEntity inviteEntity) {
         if (inviteEntity.isServiceType()) {
             UserEntity userEntity = inviteEntity.mapToUserEntity();
             ServiceEntity serviceEntity = ServiceEntity.from(Service.from());
-            if (!data.getGatewayAccountIds().isEmpty()) {
-                serviceEntity.addGatewayAccountIds(data.getGatewayAccountIds().toArray(new String[0]));
+            if (completeInviteRequest != null && !completeInviteRequest.getGatewayAccountIds().isEmpty()) {
+                serviceEntity.addGatewayAccountIds(completeInviteRequest.getGatewayAccountIds().toArray(new String[0]));
```
sfount, line 57: Having now read through the code a bit more, I wonder if it's OK to just remove the request body concept entirely and just remove this behaviour, appreciate that would involve updating a lot of tests so maybe for a separate change

sfount (approved):

src/main/java/uk/gov/pay/adminusers/model/CompleteInviteRequest.java
```diff
@@ -5,19 +5,21 @@
...
 
-    public static final String FIELD_GATEWAY_ACCOUNT_IDS = "gateway_account_ids";
-
-    @ArraySchema(schema = @Schema(implementation = String.class, example = "1,2",
+    @ArraySchema(schema = @Schema(example = "1",
             description = "gateway_account_ids that needs to be associated for the new service. Only applicable for invite type `service`"))
-    private List<String> gatewayAccountIds = new ArrayList<>();
-
-    public void setGatewayAccountIds(List<String> gatewayAccountIds) {
+    private List<String> gatewayAccountIds = Collections.emptyList();
```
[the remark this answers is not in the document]
stephencdaly: We'll be looking into this a little more when we change the registration journey on the frontend, so when we do it should be more obvious whether this is used or not. I think your suspicions are correct, so hopefully this can be made a lot simpler.

src/main/java/uk/gov/pay/adminusers/service/SelfSignupInviteCompleter.java
stephencdaly: Ah cool, I'll do it in a separate PR then. Would be good to get this point of confusion removed while its in our heads.

PR #1706, opened 2022-11-02
PP-10260 Refactor create invite resources

stephencdaly:
Refactor the resources for creating "User" and "Service" invites to use recommended Dropwizard standards for deserialising and validating the request.

Deserialise the request using a POJO with the appropriate annotations. Add `@ValidEmail` and `@ValidTelephone` number annotations and use annotations to enforce mandatory fields.


src/main/java/uk/gov/pay/adminusers/model/InviteServiceRequest.java
```diff
@@ -1,63 +1,55 @@
...
+public class InviteServiceRequest {
+    
+    private String password;
 
-public class InviteServiceRequest extends InviteRequest {
-    public static final String FIELD_PASSWORD = "password";
-    public static final String FIELD_TELEPHONE_NUMBER = "telephone_number";
-    private static final String DEFAULT_ROLE_NAME = "admin";
+    @NotEmpty
+    @ValidEmail
```
sfount, line 17: I _think_ there's a javax/ hibernate validator for email which might be able to do some of the custom validator lifting. 

I'm not sure if it validates to the same strictness as the apache commons validator but may be good enough given that we check it's a valid public sector email on top of this constraint.

```java
@NotEmpty
@Email(message = "must be a valid email address")
protected String email;
```

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -179,27 +181,21 @@ public Response getInvites(@Parameter(example = "ahq8745yq387") @QueryParam("ser
...
             responses = {
                     @ApiResponse(responseCode = "201", description = "Created",
                             content = @Content(schema = @Schema(implementation = Invite.class))),
-                    @ApiResponse(responseCode = "400", description = "Invalid payload")
+                    @ApiResponse(responseCode = "422", description = "Missing required fields or invalid values"),
+                    @ApiResponse(responseCode = "403", description = "The email is not an allowed public sector email address")
             }
     )
-    public Response createServiceInvite(JsonNode payload) {
+    public Response createServiceInvite(@Valid InviteServiceRequest inviteServiceRequest) {
```
sfount, line 191: 💯

src/main/java/uk/gov/pay/adminusers/service/ServiceInviteCreator.java
```diff
@@ -93,7 +96,7 @@ public Invite doInvite(InviteServiceRequest inviteServiceRequest) {
                         return null;
                     });
                 })
-                .orElseThrow(() -> internalServerError(format("Role [%s] not a valid role for creating a invite service request", inviteServiceRequest.getRoleName())));
+                .orElseThrow(() -> internalServerError("Unable to retrieve admin service role"));
```
sfount, line 99: This is how admin users tends to do things but explicitly throwing something called an "internalServerError" feels wrong and not very helpful from an error stack/ code readability perspective

As this ends up as a 500 with a generic message anyway I wonder if it would be OK to throw a named exception that just gets picked up by the default dropwizard error mapper

```suggestion
                .orElseThrow(() -> new AdminRoleNotFoundInDBYikes());
```

sfount (approved):
This code is so much more straight forward to read and change, anywhere manually parsing and validating json node payloads is swapped out completely changes how much you need to wade through to understand what we're trying to do

Really good stuff -- some discussion on exceptions/ validators but nothing blocking

src/main/java/uk/gov/pay/adminusers/model/InviteServiceRequest.java
stephencdaly: Ah nice, it would save on some code using that. The intention here is probably just to ensure that we don't have non-emails coming through. We also do validation on the frontend so there shouldn't be invalid emails coming through anyway.

src/main/java/uk/gov/pay/adminusers/service/ServiceInviteCreator.java
stephencdaly: I would argue that named exceptions are only useful when we need to handle them in a special way or have specific exception mapper for them so I'd prefer not to add the unnecessary boilerplate.

stephencdaly: This specific exception will disappear very soon when we stop creating a service as part of a self-sign up

sfount: Yeah I think all good especially as it's unlikely to be used elsewhere

sfount (approved):

PR #1719, opened 2022-11-09
PP-10244 Make service and role Optional on InviteEntity

stephencdaly:
An InviteEntity doesn't have a service set on it if it is for a self-registration. Make the getter for the service return an Optional to ensure we do null checks correctly.

Currently an InviteEntity always has a role set. However, this will not need to be set in the future for self-registration. It was used to set the role when creating a new test service - which was always the admin role. But will not be used when a service is no longer created as part of a self-registration so shouldn't be set on the invite in this case.


alan-gds (approved):
LGTM

PR #1724, opened 2022-11-10
PP-10038: Persist 2FA method when creating user

alan-gds:
The `/v1/api/invites/{code}/complete` endpoint now reads a `secondFactor` value from the request body and passes it on to the InviteCompleter.  The two InviteCompleters that create a new user now add the `secondFactor` to the user entity before it is persisted.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
@@ -6,7 +6,7 @@ import net.logstash.logback.marker.Markers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.adminusers.model.Invite;
-import uk.gov.pay.adminusers.model.InviteCompleteResponse;
+import uk.gov.pay.adminusers.model.CompleteInviteResponse;
 import uk.gov.pay.adminusers.persistence.dao.InviteDao;
 import uk.gov.pay.adminusers.persistence.dao.UserDao;
 import uk.gov.pay.adminusers.persistence.entity.InviteEntity;
@@ -39,7 +39,7 @@ public class NewUserExistingServiceInviteCompleter extends InviteCompleter {
 
     @Override
     @Transactional
-    public InviteCompleteResponse complete(InviteEntity inviteEntity) {
+    public CompleteInviteResponse complete(InviteEntity inviteEntity) {
         if (inviteEntity.isExpired() || Boolean.TRUE.equals(inviteEntity.isDisabled())) {
             throw inviteLockedException(inviteEntity.getCode());
         }
@@ -66,7 +66,7 @@ public class NewUserExistingServiceInviteCompleter extends InviteCompleter {
         );
 
         Invite invite = linksBuilder.addUserLink(userEntity.toUser(), inviteEntity.toInvite());
-        InviteCompleteResponse response = new InviteCompleteResponse(invite);
+        CompleteInviteResponse response = new CompleteInviteResponse(invite);
         response.setUserExternalId(userEntity.getExternalId());
 
         return response;
--- a/src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
@@ -7,17 +7,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.adminusers.model.Invite;
 import uk.gov.pay.adminusers.model.CompleteInviteResponse;
+import uk.gov.pay.adminusers.model.SecondFactorMethod;
 import uk.gov.pay.adminusers.persistence.dao.InviteDao;
 import uk.gov.pay.adminusers.persistence.dao.UserDao;
 import uk.gov.pay.adminusers.persistence.entity.InviteEntity;
 import uk.gov.pay.adminusers.persistence.entity.UserEntity;
 
+import javax.annotation.Nullable;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.conflictingEmail;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.internalServerError;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.inviteLockedException;
+import static uk.gov.pay.adminusers.service.AdminUsersExceptions.missingSecondFactorMethod;
 import static uk.gov.service.payments.logging.LoggingKeys.SERVICE_EXTERNAL_ID;
 import static uk.gov.service.payments.logging.LoggingKeys.USER_EXTERNAL_ID;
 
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
```diff
@@ -49,8 +51,12 @@ public InviteCompleteResponse complete(InviteEntity inviteEntity) {
         if (!inviteEntity.isUserType()) {
             throw internalServerError(format("Attempting to complete a 'new user, existing service' invite for an non user invite. invite-code = %s", inviteEntity.getCode()));
         }
+        if (secondFactor == null) {
+            throw internalServerError(format("Second factor method not provided when attempting to complete a 'new user, existing service' invite. invite-code = %s", inviteEntity.getCode()));
+        }
 
         UserEntity userEntity = inviteEntity.mapToUserEntity();
+        userEntity.setSecondFactor(secondFactor);
```
stephencdaly, line 59: I think it would be better to pass the second factor method into `mapToUserEntity` because otherwise it's set in two places.
It also looks like most of the tests would still pass currently if you removed this line because they are checking the method is set to `SMS`, but we default to `SMS` in `mapToUserEntity`, so you should ensure they fail if we aren't using the `SecondFactorMethod` that comes through in the request.

src/main/java/uk/gov/pay/adminusers/service/NewUserExistingServiceInviteCompleter.java
```diff
@@ -49,8 +51,12 @@ public InviteCompleteResponse complete(InviteEntity inviteEntity) {
         if (!inviteEntity.isUserType()) {
             throw internalServerError(format("Attempting to complete a 'new user, existing service' invite for an non user invite. invite-code = %s", inviteEntity.getCode()));
         }
+        if (secondFactor == null) {
+            throw internalServerError(format("Second factor method not provided when attempting to complete a 'new user, existing service' invite. invite-code = %s", inviteEntity.getCode()));
```
stephencdaly, line 55: I think this error should be a bad request error, and return a 400 status code because it's failing because something is missing from the request body, rather than anything internal to the app that's gone wrong.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceServiceCompleteIT.java
```diff
@@ -20,7 +20,7 @@ class InviteResourceServiceCompleteIT extends IntegrationTest {
     public static final String INVITES_RESOURCE_URL = "/v1/api/invites";
 
     @Test
-    void shouldReturn200withDisabledInviteLinkingToCreatedUser_WhenPassedAValidInviteCode() {
+    void shouldReturn200withDisabledInviteLinkingToCreatedUser_WhenPassedAValidInviteCode() throws Exception {
```
stephencdaly, line 23: Could we add a test case to check that a 422 error is returned if the `second_factor` provided isn't an allowed value?

src/test/java/uk/gov/pay/adminusers/service/ExistingUserInviteCompleterTest.java
```diff
@@ -86,6 +86,29 @@ public void shouldSuccess_whenSubscribingAServiceToAnExistingUser_forValidInvite
         assertThat(persistedUser.getValue().getServicesRole(service.getExternalId()).isPresent(), is(true));
     }
 
+    @Test
+    public void shouldSuccess_whenSubscribingAServiceToAnExistingUser_forValidInvite_with2FAMethod() {
```
stephencdaly, line 90: In this case completing the invite does not create a user as it already exists and we're not doing anything with the second factor method passed in, so I don't think this is testing anything. I think it's just passing because the user we return from the DAO already has SMS set as the `second_factor` method?

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceServiceCompleteIT.java
```diff
@@ -30,8 +30,12 @@ void shouldReturn200withDisabledInviteLinkingToCreatedUser_WhenPassedAValidInvit
                 .withEmail(email)
                 .withPassword(password)
                 .insertServiceInvite();
+
+        Map<String, String> payload = Map.of("second_factor", "SMS");
```
stephencdaly, line 34: Possibly it's better if all the tests check setting the `second_factor` method to "APP" as it is less likely to be a default if we were setting a default anywhere?

stephencdaly (changes requested):

src/test/java/uk/gov/pay/adminusers/service/ExistingUserInviteCompleterTest.java
alan-gds: Yeah, you're right that this is testing nothing as it stands.  I do think we should represent the situation with a (better) test though - what do you think the behaviour ought to be if we somehow receive an 'existing user' invite completion request with a (redundant) 2FA method in the body?

stephencdaly: I'd say as long as we're testing it works fine without a `second_factor` method provided we're fine, which I think we do in other tests. If we wanted to be really thorough we could check that the `second_factor` is ignored and not updated on the user but that doesn't strike me as particularly necessary.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceServiceCompleteIT.java
alan-gds: It turns out DropWizard (or so I presume) returns a 400 when it tries to match the argument value to the enum and fails, so we can't easily have a 422, but I've added a test to check for the 400.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceServiceCompleteIT.java
alan-gds: Agreed.  I thought about doing this in the first place but then didn't.

src/main/java/uk/gov/pay/adminusers/service/SelfSignupInviteCompleter.java
```diff
@@ -42,13 +45,16 @@ public SelfSignupInviteCompleter(InviteDao inviteDao, UserDao userDao, ServiceDa
...
-    public InviteCompleteResponse complete(InviteEntity inviteEntity) {
+    public CompleteInviteResponse complete(InviteEntity inviteEntity, @Nullable SecondFactorMethod secondFactor) {
         if (inviteEntity.isExpired() || inviteEntity.isDisabled()) {
             throw inviteLockedException(inviteEntity.getCode());
         }
         if (userDao.findByEmail(inviteEntity.getEmail()).isPresent()) {
             throw conflictingEmail(inviteEntity.getEmail());
         }
+        if (secondFactor == null) {
+            throw internalServerError(format("Second factor method not provided when attempting to complete a 'new user, existing service' invite. invite-code = %s", inviteEntity.getCode()));
```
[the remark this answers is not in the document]
alan-gds: This was probably a copy/paste error, but you're right that it makes more sense to harmonise the messages anyway.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceServiceCompleteIT.java
```diff
@@ -57,7 +62,30 @@ void shouldReturn200withDisabledInviteLinkingToCreatedUser_WhenPassedAValidInvit
...
+        String telephoneNumber = "+447700900000";
+        String password = "valid_password";
+
+        Map<String, String> payload = Map.of("second_factor", "BAD_VALUE");
+
+        String inviteCode = inviteDbFixture(databaseHelper)
+                .withTelephoneNumber(telephoneNumber)
+                .withEmail(email)
+                .withPassword(password)
+                .insertServiceInvite();
```
stephencdaly, line 77: This test shouldn't need to insert an invite into the database as it shouldn't get that far. Maybe worth removing this to avoid any confusion.

stephencdaly (approved):
LGTM!

PR #1727, opened 2022-11-11
PP-10049 Use ValidationException from pay-java-commons

stephencdaly:
So that we can use shared validation code from pay-java-commons, use ValidationException from pay-java-commons.

Update ValidationExceptionMapper to map the pay-java-commons exception to a response of the same format as is currently returned.


src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -13,14 +13,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.adminusers.exception.ServiceNotFoundException;
-import uk.gov.pay.adminusers.exception.ValidationException;
 import uk.gov.pay.adminusers.model.*;
```
kbottla, line 16: Not related to this PR but noticed a few wild card imports `import uk.gov.pay.adminusers.model.*. Would be good if they can be removed in future PR

kbottla (approved):

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
stephencdaly: Good spot, will remove in my next PR

PR #1728, opened 2022-11-11
PP-10049 Add patch request to update invite

stephencdaly:
Add a patch request endpoint to update an invite. This accepts replace operations to update the "password" and "telephone_number".

Uses shared classes from pay-java-commons to do the patch request validation.

We validate that the telephone number is a valid phone number. For the password we do not enforce any rules other than a string is provided, as our password rules are only currently enforced by the frontend.

## Review notes

Hiding whitespace in the diff will help as indentation changed by adding nested classes in the tests


kbottla (approved):
LGTM!

PR #1745, opened 2022-11-17
PP-10329: Add send-otp endpoint to InviteResource

alan-gds:
New endpoint is `/v1/api/invites/{code}/send-otp`

This will later replace `/v1/api/invites/{code}/otp/generate` and `/v1/api/invites/otp/resend`

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
@@ -129,6 +129,28 @@ public class InviteResource {
                 .collect(Collectors.toList());
         return inviteService.updateInvite(inviteCode, updateRequests);
     }
+    
+    @POST
+    @Path("/v1/api/invites/{code}/send-otp")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Sends otp verification code to the phone number registered in the invite",
+            responses = {
+                    @ApiResponse(responseCode = "204", description = "No content"),
+                    @ApiResponse(responseCode = "400", description = "Invalid payload"),
+                    @ApiResponse(responseCode = "404", description = "Not found"),
+                    @ApiResponse(responseCode = "412", description = "Precondition failed")
+            }
+    )
+    public void sendOtp(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode) {
+        LOGGER.info("Invite send OTP POST request for code - [ {} ]", inviteCode);
+
+        if (isNotBlank(inviteCode) && inviteCode.length() > MAX_LENGTH_CODE) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND);
+        }
+
+        inviteService.sendOtp(inviteCode);
+    }
 
     @POST
     @Path("/v1/api/invites/{code}/complete")
--- a/src/main/java/uk/gov/pay/adminusers/service/InviteService.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/InviteService.java
@@ -32,6 +32,7 @@ import java.util.Optional;
 import static uk.gov.pay.adminusers.resources.InviteRequestValidator.FIELD_PASSWORD;
 import static uk.gov.pay.adminusers.resources.InviteRequestValidator.FIELD_TELEPHONE_NUMBER;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.invalidOtpAuthCodeInviteException;
+import static uk.gov.pay.adminusers.service.AdminUsersExceptions.inviteDoesNotHaveTelephoneNumberError;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.inviteLockedException;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.missingSecondFactorMethod;
 import static uk.gov.pay.adminusers.service.AdminUsersExceptions.notFoundInviteException;
@@ -91,6 +92,26 @@ public class InviteService {
             }
         } else {
             throw notFoundInviteException(resendOtpRequest.getCode());
+        }
+    }
+
+    @Transactional
+    public void sendOtp(String inviteCode) {
+        InviteEntity invite = findInvite(inviteCode).orElseThrow(() -> notFoundInviteException(inviteCode));
+
+        if (invite.getTelephoneNumber() == null) {
+            throw inviteDoesNotHaveTelephoneNumberError(inviteCode);
+        }
+
+        int passCode = secondFactorAuthenticator.newPassCode(invite.getOtpKey());
+        String formattedPasscode = String.format(Locale.ENGLISH, SIX_DIGITS_WITH_LEADING_ZEROS, passCode);
+
+        OtpNotifySmsTemplateId templateId = invite.getService().isPresent() ? CREATE_USER_IN_RESPONSE_TO_INVITATION_TO_SERVICE : SELF_INITIATED_CREATE_NEW_USER_AND_SERVICE;
+        try {
+            String notificationId = notificationService.sendSecondFactorPasscodeSms(invite.getTelephoneNumber(), formattedPasscode, templateId);
+            LOGGER.info("sent 2FA token successfully for invite code [{}], notification id [{}]", invite.getCode(), notificationId);
+        } catch (Exception e) {
+            LOGGER.error(String.format("error sending 2FA token for invite code [%s]", invite.getCode()), e);
         }
     }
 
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -129,6 +131,27 @@ public Invite updateInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv")
...
+    @Path("/v1/api/invites/{code}/send-otp")
+    @Produces(APPLICATION_JSON)
+    public Response sendOtp(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode) {
+        LOGGER.info("Invite send OTP POST request for code - [ {} ]", inviteCode);
+
+        if (isNotBlank(inviteCode) && inviteCode.length() > MAX_LENGTH_CODE) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND);
+        }
+
+        return inviteService.findInvite(inviteCode).map(inviteEntity -> {
```
stephencdaly, line 145: Finding the invite and checking whether there's a telephone number is probably better happening in the InviteService. We like to keep the resource as slim as possible, and have it mainly detailing with parsing the request and constructing the response.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -129,6 +131,27 @@ public Invite updateInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv")
...
+            if (inviteEntity.getTelephoneNumber() == null) {
+                return Response.status(PRECONDITION_FAILED).build();
```
stephencdaly, line 147: The error response needs to have a suitable error message with it. If we make the resource method return `void`, we would return an error response by instead throwing a `WebApplicationException`. We should add a suitable exception builder in `AdminusersExceptions` for this.

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -93,6 +93,18 @@ public void reGenerateOtp(ResendOtpRequest resendOtpRequest) {
             throw notFoundInviteException(resendOtpRequest.getCode());
         }
     }
+    
+    @Transactional
+    public void sendOtp(InviteEntity invite) {
+        OtpNotifySmsTemplateId templateId = (invite.getService().isPresent()) ? CREATE_USER_IN_RESPONSE_TO_INVITATION_TO_SERVICE : SELF_INITIATED_CREATE_NEW_USER_AND_SERVICE;
+
+        try {
+            String notificationId = notificationService.sendSecondFactorPasscodeSms(invite.getTelephoneNumber(), invite.getPassword(), templateId);
```
stephencdaly, line 102: This doesn't look right, we shouldn't be sending the password in the SMS message! We need to generate an otp code from the `otp_key` on the invite. Take a look at the existing code to see how we do this.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java
```diff
@@ -183,6 +185,58 @@ void validateOtp_shouldFailWith401_whenInvalidOtp() throws Exception {
...
+                .insertInviteToAddUserToService();
+
+        givenSetup()
+                .when()
+                .contentType(JSON)
+                .post(format(INVITES_SEND_OTP_RESOURCE_URL, code))
+                .then()
+                .statusCode(NO_CONTENT.getStatusCode());
+        
+        // TODO - validate that notification is sent?
```
stephencdaly, line 202: Yes it would be good if the integration test verified that a mocked endpoint on Notify is called. However, we currently start a DropWizard app to mock out Notify rather than using WireMock so this would be difficult. Perhaps we can leave this as is for now, and then in another PR look into how we could use WireMock to mock out Notify so we can do the verification.

stephencdaly (changes requested):

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -129,6 +131,27 @@ public Invite updateInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv")
                 .collect(Collectors.toList());
         return inviteService.updateInvite(inviteCode, updateRequests);
     }
+    
+    @POST
+    @Path("/v1/api/invites/{code}/send-otp")
+    @Produces(APPLICATION_JSON)
+    public Response sendOtp(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode) {
```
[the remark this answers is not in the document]
alan-gds: Fair enough; now sorted.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
alan-gds: Yeah, I wasn't keen doing this stuff in the resource, particularly checking the telephone number.  Your way makes a lot more sense.

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
alan-gds: Yeah, I did think this was weird.  I think I did look at existing code but misinterpreted it.  I've now sorted it out.

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java
alan-gds: Agreed.  Should I remove the TODO now, or is it okay to leave it as we know we'll be following this up immediately?

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -94,6 +95,32 @@ public void reGenerateOtp(ResendOtpRequest resendOtpRequest) {
         }
     }
 
+    @Transactional
+    public void sendOtp(String inviteCode) {
+        var inviteMaybe = findInvite(inviteCode);
+        if (inviteMaybe.isPresent()) {
```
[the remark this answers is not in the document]
alan-gds: Ah yeah, that's definitely better.

src/main/java/uk/gov/pay/adminusers/resources/InviteResource.java
```diff
@@ -129,6 +129,19 @@ public Invite updateInvite(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv")
                 .collect(Collectors.toList());
         return inviteService.updateInvite(inviteCode, updateRequests);
     }
+    
+    @POST
+    @Path("/v1/api/invites/{code}/send-otp")
+    @Produces(APPLICATION_JSON)
+    public void sendOtp(@Parameter(example = "d02jddeib0lqpsir28fbskg9v0rv") @PathParam("code") String inviteCode) {
```
stephencdaly, line 136: we should add some `@ApiResponse` annotations to this so we document the possible responses correctly in the OpenApi file

src/main/java/uk/gov/pay/adminusers/service/InviteService.java
```diff
@@ -94,6 +95,26 @@ public void reGenerateOtp(ResendOtpRequest resendOtpRequest) {
...
+        InviteEntity invite = findInvite(inviteCode).orElseThrow(() -> notFoundInviteException(inviteCode));
+
+        if (invite.getTelephoneNumber() == null) {
+            throw inviteDoesNotHaveTelephoneNumberError(inviteCode);
+        }
+
+        int passCode = secondFactorAuthenticator.newPassCode(invite.getOtpKey());
+        String formattedPasscode = String.format(Locale.ENGLISH, SIX_DIGITS_WITH_LEADING_ZEROS, passCode);
+
+        OtpNotifySmsTemplateId templateId = (invite.getService().isPresent()) ? CREATE_USER_IN_RESPONSE_TO_INVITATION_TO_SERVICE : SELF_INITIATED_CREATE_NEW_USER_AND_SERVICE;
```
stephencdaly, line 109: I think there are some un-needed brackets here

stephencdaly (changes requested):
Looks good, just need to add some swagger annotations so the OpenApi file includes the responses

src/test/java/uk/gov/pay/adminusers/resources/InviteResourceOtpIT.java
alan-gds: Raised as https://payments-platform.atlassian.net/browse/PP-10368; please have a look and see if I should change anything.

stephencdaly (approved):
LGTM!

PR #1748, opened 2022-11-18
PP-10199 Add GoLiveStage - CHOSEN_GOV_BANKING_MOTO

james-peacock-gds:
## WHAT YOU DID
- To support new page in onboarding process
- Added GOV_BANKING_MOTO_OPTION_COMPLETED value to enum GoLiveStage
- Updated associated test and api spec


src/main/java/uk/gov/pay/adminusers/model/GoLiveStage.java
```diff
@@ -9,6 +9,7 @@ public enum GoLiveStage {
     CHOSEN_PSP_SMARTPAY,
     CHOSEN_PSP_EPDQ,
     CHOSEN_PSP_GOV_BANKING_WORLDPAY,
+    CHOSEN_GOV_BANKING_MOTO,
```
SandorArpa, line 12: I think it should be `CHOSEN_PSP_GOV_BANKING_MOTO`

SandorArpa (changes requested):

src/main/java/uk/gov/pay/adminusers/model/GoLiveStage.java
james-peacock-gds: thanks Sandor, that is a good point.  I was going from what Krishna specified in the ticket, but I see what you mean - I had made the assumption that for the others, e.g. CHOSEN_PSP_SMARTPAY, it's the PSP that is being chosen, whereas in this particular case it is the MOTO option that is being chosen rather than the PSP specifically... but I might have misunderstood. I will check with Krishna on Monday.

SandorArpa: 👍

james-peacock-gds: @kbottla please can you confirm if the variable name should be altered? Thanks

kbottla: As discussed, let's rename it to `GOV_BANKING_MOTO_OPTION_COMPLETED` to make it clear that the service has completed the step.

SandorArpa (approved):
LGTM

PR #1761, opened 2022-11-25
PP-10052 Add pact state for invite for existing user

stephencdaly:
Add pact state so we can add a pact to check that is possible to call /v1/api/invites/${inviteCode}/complete without a request body for an invite for an existing user.


james-peacock-gds (approved):
LGTM

PR #1789, opened 2022-12-13
PP-10338: Add `isInviteToJoinService` to `Invite` object

alan-gds:
Endpoints which return an invite will now include this flag, which will soon be used instead of the `type` field, so that we can delete the latter.


stephencdaly (commented):
Can we add some tests for this please?

stephencdaly (approved):

PR #1798, opened 2022-12-19
PP-10338: Remove type field from invites

alan-gds:
Instead of using the `type` field from invite responses, selfservice now uses `is_invite_to_join_service`, so `type` is no longer needed.  This also means the `InviteType` enum is redundant, so it has been removed.


stephencdaly (approved):
Nice

PR #1799, opened 2022-12-19
PP-10338: Remove `invite.type` column from database

alan-gds:
Invite types are no longer used, so the column is now redundant.


stephencdaly (approved):
LGTM!

PR #1929, opened 2023-03-21
ZD-5201821: Allow idoxgroup.com as a user domain

jfharden:
## WHAT YOU DID
Allow users with the domain idoxgroup.com to be created


sfount:
(x-post) I think the ticket can be resolved without an additional exception (through invites), should be OK to close this unless there are further needs

PR #2069, opened 2023-07-17
PP-11265 add unique index to lower email on the users table

SandorArpa:
## WHAT YOU DID

The current sql unique constraint on the email column in the users table is not case insensitive. It should be.

Unique constraint lower(email) is not valid PSQL script. It was either unique index or installing citext extension and altering the column type. citext can only be installed by superuser and cannot be done through our usual liquibase script.

- Add unique index to lower(email).

### DIFF
--- a/src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
+++ b/src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
@@ -23,12 +23,14 @@ import static java.lang.String.valueOf;
 import static java.util.stream.Collectors.toUnmodifiableList;
 import static org.apache.commons.lang3.RandomUtils.nextInt;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.emptyOrNullString;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/adminusers/persistence/dao/UserDaoIT.java
```diff
@@ -349,4 +351,46 @@ public void shouldNotFindAnyUser() {
...
+        Optional<ServiceEntity> serviceEntity = serviceDao.findById(serviceId);
+        RoleEntity roleEntity = roleDao.findByRoleName(role.getName()).get();
+
+        ServiceRoleEntity serviceRoleEntity = new ServiceRoleEntity(serviceEntity.get(), roleEntity);
+        serviceRoleEntity.setUser(userEntity);
+
+        userEntity.addServiceRole(serviceRoleEntity);
+
+        var thrown = assertThrows(javax.persistence.RollbackException.class, () -> userDao.persist(userEntity));
+        assertThat(thrown.getMessage(), containsString("ERROR: duplicate key value violates unique constraint \"lower_case_email_index\""));
```
kbottla, line 394: Is the plan to handle this error later in the service/resource layer and return 409?

kbottla (approved):

PR #2079, opened 2023-07-25
PP-11283 drop not null constraint

SandorArpa:
## WHAT YOU DID
We are removing username column. Before we can drop the column, we need to refactor the code, and that's not possible without dropping the not null constraint

- drop not null constrain for username in users table


AbdirizakIdris (approved):
LGTM

PR #2087, opened 2023-07-31
PP-11283 Remove remaining usernames from adminusers

AbdirizakIdris:
## WHAT
- Removed remaining uses of username from endpoints and replaced it to return email


stephencdaly (approved):
LGTM!

PR #2089, opened 2023-07-31
PP-11283 Drop username column

AbdirizakIdris:
## WHAT
- Dropped ``username`` column from ``users`` table


SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #2091, opened 2023-08-01
PP-11327 refactor authentication to use email

SandorArpa:
## WHAT YOU DID

We removed username from the users DB. To avoid confusion, we're going to use email instead of username when authenticating a user.

- refactor authentication to use email


stephencdaly (approved):
LGTM!

PR #2094, opened 2023-08-03
PP-11292: Update pact broker host

oswaldquek:


jonathan-ryding (approved):

.github/workflows/_run-pact-provider-tests.yml
```diff
@@ -58,6 +58,6 @@ jobs:
           -DPACT_CONSUMER_TAG="${{ inputs.consumer_tag }}" \
           -Dpact.provider.version="${{ steps.get-provider-sha.outputs.provider-sha }}" \
           -Dpact.verifier.publishResults=true \
-          -DPACT_BROKER_HOST=pay-pact-broker.cloudapps.digital \
+          -DPACT_BROKER_HOST=pact-broker.deploy.payments.service.gov.uk \
```
sfount, line 61: Appreciate this likely won't need to change again for the foreseeable so probably not worth it now but I wonder if it would be possible to configure the various actions to use [variables](https://docs.github.com/en/actions/learn-github-actions/variables). 

Something like 
```
-DPACT_BROKER_HOST="${{ vars.govuk_pay_pact_broker_host }}" \
```

That var could then be defined once by the organisation and shared across any of the repos that use it -- I think they'd then just run with the latest value thats set.

nlsteers: Yeah that would be really useful wouldn't it

PR #2147, opened 2023-09-13
BAU Return success response when agreement exists

stephencdaly:
When a request was made to create either a GovUkPayAgreement or StripeAgreement was made, we were returning a 409 error if one already existed.

Instead return a 200. For creating a GovUkPayAgreement, also return the agreement in the response.

This is so that we can recover from the scenario where the end user submits the admin tool page to agree to the agreement, but there is another error, which we have seen happen when trying to create the Zendesk ticket.

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
@@ -411,8 +411,8 @@ public class ServiceResource {
             summary = "Record acceptance of Stripe terms",
             description = "Records that a GOV.UK Pay agreement has been accepted for the service.",
             responses = {
+                    @ApiResponse(responseCode = "200", description = "Stripe agreement already created"),
                     @ApiResponse(responseCode = "201", description = "Created"),
-                    @ApiResponse(responseCode = "409", description = "Conflict. Another stripe agreement already exists for the service"),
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -421,6 +421,12 @@ public Response createStripeAgreement(@Parameter(example = "7d19aff33f8948deb97e
                                           @NotNull @Valid StripeAgreementRequest stripeAgreementRequest) throws UnknownHostException {
         LOGGER.info("Create stripe agreement POST request - [ {} ]", stripeAgreementRequest.toString());
 
+        Optional<StripeAgreement> maybeStripeAgreement = stripeAgreementService.findStripeAgreementByServiceId(serviceExternalId);
+        if (maybeStripeAgreement.isPresent()) {
```
kbottla, line 425: Although rare, what if a different user resumes the journey? We won't be storing details of the current user that continues the go-live request in this case.

stephencdaly: Someone has agreed to the terms - and they have essentially completed the go-live at this point. The only reason an agreement acceptance will have been stored and they're able to accept again is if there's been an error submitting the page.

kbottla (approved):

PR #2185, opened 2023-10-20
PP-11652 Upgrade from Dropwizard 3.0.1 to Dropwizard 3.0.2

alexbishop1:
• Upgrade Dropwizard from 3.0.1 to 3.0.2
• Upgrade WireMock from 2.35.1 to 3.2.0
• Use more Maven BOMs for dependencies
• Separate dependencies into those imported from BOMs and others


SandorArpa (approved):
LOKTM

PR #2191, opened 2023-10-25
BAU — Small tweaks to POM and Dependabot config

alexbishop1:
- Use Dropwizard version property so we upgrade Dropwizard and Dropwizard Testing Support for JUnit 4.x in lockstep
- Use Testcontainers BOM
- Stop using EclipseLink version property because it’s only used in one place
- Remove unused Guice version property
- Tell Dependabot not to try to upgrade us to Guice 7.0.0, which migrates from Java EE to Jakarta EE and is never going to be a just-merge-it upgrade


SandorArpa (approved):
LGTM

PR #2198, opened 2023-10-26
PP-11205 Add service method to expunge data

kbottla:
## WHAT YOU DID
- Added service method to delete historical data.
- Currently includes
    - Users not attached to any service
    - Forgotten passwords
    - and Invites


oswaldquek (approved):
LGTM

PR #2221, opened 2023-11-03
ZD5545187: Add the domain ukaea.uk to allow list

jonathan-ryding:
Add the domain ukaea.uk to allow list so that users with email addresses ending in ukaea.uk can create test accounts.


SandorArpa (approved):
LGTM

PR #2224, opened 2023-11-06
PP-11205 Archive services

kbottla:
## WHAT YOU DID
- Archive services without any transactions in the last 7 years (or based on config number of days).
- Criteria for archiving service is 
  - last transaction (for any linked gateway account) and is older than 7 years
  -  OR created date is older than 7 years

- It is possible that a service doesn't have any transaction and created_date is also null. These will be dealt with separately.


stephencdaly (approved):
LGTM!

PR #2229, opened 2023-11-07
PP-11205 Detach users from service

kbottla:
## WHAT YOU DID
- Removes users from service when service is archived.


SandorArpa (approved):
LGTM

PR #2241, opened 2023-11-10
PP-11355 allow apostrophes and hyphens when searching by service name

SandorArpa:
## WHAT YOU DID

Some services - such as King’s Academy have an apostrophe in their name.

- update regex to allow apostrophes and hyphen in service search payload

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
@@ -100,7 +100,7 @@ public class ServiceRequestValidator {
     }
 
     public Optional<Errors> validateSearchRequest(ServiceSearchRequest request) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/adminusers/resources/ServiceRequestValidator.java
```diff
@@ -100,7 +100,7 @@ private void validateMerchantEmail(JsonNode payload) throws ValidationException
     }
 
     public Optional<Errors> validateSearchRequest(ServiceSearchRequest request) {
-        var allowedChars = Pattern.compile("^[0-9A-Za-z\\s]+$");
+        var allowedChars = Pattern.compile("^[0-9A-Za-z'\\-\\s]+$");
```
alexbishop1, line 103: While you’re here, you could make this into a constant so we don’t compile the regex every time.

alexbishop1 (approved):
Looks fine. Make sure NAXSI is happy with it.

PR #2275, opened 2023-11-24
PP-11812 Log when users are removed from service

kbottla:
## WHAT YOU DID
- Log user IDs when users are removed from service after a service is archived. This will be useful when a service is accidentally archived and we have to reinstate users


SandorArpa (approved):
LGTM

PR #2492, opened 2024-05-07
PP-12570 Run pact tests in GitHub Actions for PRs

kbottla:
## WHAT YOU DID
- Run pact tests in GitHub actions for PRs


jfharden:
It's better not to prefix run_tests with an underscore in the filename. The only things with that naming convention are workflows which are purely reusable and never run in their own right

jfharden:
> It's better not to prefix run_tests with an underscore in the filename. The only things with that naming convention are workflows which are purely reusable and never run in their own right

Scratch that, I just saw you added pr.yml and called it from there 🤦

.github/workflows/_run-tests.yml
```diff
@@ -0,0 +1,54 @@
+name: Github Actions Tests
+
+on:
+  workflow_call:
+    secrets:
+      pact_broker_username:
+        required: true
+        description: required if `publish_pacts` is `true`
```
jfharden, line 8: There's isn't a separate public_pacts, I think you can remove this description (and from the one below)

kbottla: Removed

.github/workflows/pr.yml
```diff
@@ -0,0 +1,16 @@
+name: PR
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency: adminusers-pr
```
jfharden, line 9: Do we want this concurrency? Having it means that only a single PR will be able to run tests at once which might be quite painful for developers

kbottla: might be. Dependabot PRs open at once which might use loads of resources. We can remove it if it causes any issues for developers but it is very rare

jfharden: On the plus side, the resources are all GitHubs, not ours. Concourse running the e2e tests is hardly doing anything since it runs in CodeBuild, so it's mostly just polling waiting for the CodeBuild to finish.

We never had concurrency on the PR tests before and I'm quite reluctant for us to introduce it since it's an extra level of complexity, and I think an unnecessary slowing down of testing.

kbottla: 👍🏼 removed

jfharden (approved):

PR #2525, opened 2024-05-23
BAU ignore org.eclipse.persistence.jpa

SandorArpa:
## WHAT YOU DID

org.eclipse.persistence.jpa 3.x and above only works with Jakarta EE and not Java EE

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,10 @@ updates:
         # Guice 7.x only works with Jakarta EE and not Java EE
         versions:
           - ">= 6"
+      - dependency-name: "org.eclipse.persistence:org.eclipse.persistence.jpa"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -27,6 +27,10 @@ updates:
         # Guice 7.x only works with Jakarta EE and not Java EE
         versions:
           - ">= 6"
+      - dependency-name: "org.eclipse.persistence:org.eclipse.persistence.jpa"
+        # Version 4 is based on Jakarta EE
```
alexbishop1, line 31: So is version 3. For consistency with the other comments, can we make this say “Jakarta Persistence 3.x only works with Jakarta EE and not Java EE”?

alexbishop1 (changes requested):
Good spot!

PR #2568, opened 2024-06-20
PP-12780: Remove arm dockerfile

kerrr:
We no longer use the arm version of the docker, all builds use the main dockerfile. Therefor we can remove this one.

* Delete m1/arm64.Dockerfile


jfharden (approved):

jfharden (changes requested):
Oh actually, you need to update the github workflow that tests for it being a manifest

jfharden (approved):

PR #2588, opened 2024-07-03
PP-11723 upgrade guice to v6

SandorArpa:
## WHAT YOU DID

Upgrade project to use Guice v6

Add Transactional annotation to RoleDao. It was missing and Guice 6 is a stricter implementation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
             <dependency>
                 <groupId>com.google.inject</groupId>
                 <artifactId>guice-bom</artifactId>
-                <version>5.1.0</version>
+                <version>6.0.0</version>
                 <type>pom</type>
                 <scope>import</scope>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -523,6 +546,13 @@
             </plugin>
         </plugins>
     </build>
+    <repositories>
+        <repository>
+            <id>maven_central</id>
+            <name>Maven Central</name>
+            <url>https://repo.maven.apache.org/maven2/</url>
+        </repository>
+    </repositories>
```
alexbishop1, line 555: Do we need this? Isn’t it the default?

alexbishop1 (changes requested):
Can we also update the `dependabot.yml` file?

alexbishop1 (approved):

PR #2620, opened 2024-07-31
BAU: Add Bolton Sixth Form College

nataliecarey:
This will allow users with email addresses ending with bolton-sfc.ac.uk to create test accounts.


hjvoid (approved):
LGTM 🤠

PR #2653, opened 2024-09-06
PP-13030: Use CreateServiceRequest pojo instead of JsonNode

oswaldquek:
Following the general pattern in most of our Dropwizard codebases, we want to avoid JsonNode as the POST request body where possible so we can advantage of using javax Validation constraints.

It is also generally agreed that using hard coded URL strings makes for easier-to-read code.

Added a few more test cases in ServiceResourceCreateIT to reflect the pact tests between selfservice and adminusers:
* [a valid create service request with empty object](https://github.com/alphagov/pay-selfservice/blob/b2053a1049bc28813e8e7363245ad546ef19b675/test/pact/adminusers-client/service/create-service.pact.test.js#L50)
* [a valid create service request with gateway account ids](https://github.com/alphagov/pay-selfservice/blob/b2053a1049bc28813e8e7363245ad546ef19b675/test/pact/adminusers-client/service/create-service.pact.test.js#L85C31-L85C86)
* [a valid create service request with service name](https://github.com/alphagov/pay-selfservice/blob/b2053a1049bc28813e8e7363245ad546ef19b675/test/pact/adminusers-client/service/create-service.pact.test.js#L124)

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/ServiceNamesDeserializer.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/ServiceNamesDeserializer.java
@@ -7,17 +7,15 @@ import com.fasterxml.jackson.databind.JsonDeserializer;
 import uk.gov.service.payments.commons.model.SupportedLanguage;
 
 import java.io.IOException;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
+
+import static java.util.stream.Collectors.toUnmodifiableMap;
 
 public class ServiceNamesDeserializer extends JsonDeserializer<Map<SupportedLanguage, String>> {
     @Override
     public Map<SupportedLanguage, String> deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
-        Map<SupportedLanguage, String> supportedLanguageToServiceName = new HashMap<>();
-        jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, String>>() {})
-                .forEach((key, value) -> 
-                        supportedLanguageToServiceName.put(SupportedLanguage.fromIso639AlphaTwoCode(key), value));
-        return Collections.unmodifiableMap(supportedLanguageToServiceName);
+        return jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, String>>() {})
+                .entrySet().stream()
+                .collect(toUnmodifiableMap(entry -> SupportedLanguage.fromIso639AlphaTwoCode(entry.getKey()), Map.Entry::getValue));
     }
 }
--- a/src/main/java/uk/gov/pay/adminusers/model/ServiceNamesDeserializer.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/ServiceNamesDeserializer.java
@@ -1,0 +1,23 @@
+package uk.gov.pay.adminusers.model;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import uk.gov.service.payments.commons.model.SupportedLanguage;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ServiceNamesDeserializer extends JsonDeserializer<Map<SupportedLanguage, String>> {
+    @Override
+    public Map<SupportedLanguage, String> deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        Map<SupportedLanguage, String> supportedLanguageToServiceName = new HashMap<>();
+        jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, String>>() {})
+                .forEach((key, value) -> 
+                        supportedLanguageToServiceName.put(SupportedLanguage.fromIso639AlphaTwoCode(key), value));
+        return Collections.unmodifiableMap(supportedLanguageToServiceName);
+    }
+}
--- a/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
+++ b/src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
@@ -13,6 +13,7 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.adminusers.exception.ServiceNotFoundException;
+import uk.gov.pay.adminusers.model.CreateServiceRequest;
 import uk.gov.pay.adminusers.model.GovUkPayAgreement;
 import uk.gov.pay.adminusers.model.SearchServicesResponse;
 import uk.gov.pay.adminusers.model.Service;
@@ -33,7 +34,6 @@ import uk.gov.pay.adminusers.service.ServiceServicesFactory;
 import uk.gov.pay.adminusers.service.StripeAgreementService;
 import uk.gov.pay.adminusers.utils.Errors;
 import uk.gov.service.payments.commons.api.exception.ValidationException;
-import uk.gov.service.payments.commons.model.SupportedLanguage;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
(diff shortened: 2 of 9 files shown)


src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateIT.java
```diff
@@ -1,37 +1,117 @@
...
+    
+    @Test
+    void return_bad_request_when_invalid_supported_language_provided() {
+        givenSetup()
+                .when()
+                .contentType(JSON)
+                .body(Map.of("service_name", Map.of("fr", "Service name")))
+                .post("v1/api/services")
+                .then()
+                .statusCode(400)
```
oswaldquek, line 113: Returning a 422 might be more correct here, but getting Jersey to do this from within a custom deserialiser is very hard to do. We return a [400 in other places](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/charge/resource/ChargesApiResourceValidationTest.java#L199) so I think this is acceptable.

src/main/java/uk/gov/pay/adminusers/model/ServiceNamesDeserializer.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Map;
+
+public class ServiceNamesDeserializer extends JsonDeserializer<Map<SupportedLanguage, String>> {
+    @Override
+    public Map<SupportedLanguage, String> deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        Map<SupportedLanguage, String> supportedLanguageToServiceName = new HashMap<>();
+        jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, String>>() {})
+                .forEach((key, value) -> 
+                        supportedLanguageToServiceName.put(SupportedLanguage.fromIso639AlphaTwoCode(key), value));
+        return Collections.unmodifiableMap(supportedLanguageToServiceName);
```
alexbishop1, line 21:
```java
return jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, String>>() {})
        .entrySet().stream()
        .collect(toUnmodifiableMap(entry ->
                SupportedLanguage.fromIso639AlphaTwoCode(entry.getKey()), Map.Entry::getValue));
```
😀

oswaldquek: Nice, fixed.

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -227,28 +220,21 @@ public Response searchServices(JsonNode payload) {
                     @ApiResponse(responseCode = "201", description = "Created",
                             content = @Content(schema = @Schema(implementation = Service.class))),
                     @ApiResponse(responseCode = "409", description = "Gateway account IDs provided has already been assigned to another service"),
+                    @ApiResponse(responseCode = "400", description = "Service names must be one of 'en' (English) or 'cy' (Welsh)"),
```
alexbishop1, line 223: I suppose really it’s the service name _key_ that needs to be ‘en’ or ‘cy’.

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
```diff
@@ -227,28 +220,21 @@ public Response searchServices(JsonNode payload) {
...
-        List<String> gatewayAccountIds = extractGatewayAccountIds(payload);
-        Map<SupportedLanguage, String> serviceNameVariants = getServiceNameVariants(payload);
+    public Response createService(@Valid CreateServiceRequest createServiceRequest) {
+        LOGGER.info("Create Service POST request - [ {} ]", createServiceRequest);
 
-        Service service = serviceServicesFactory.serviceCreator().doCreate(gatewayAccountIds, serviceNameVariants);
+        var nullableCreateServiceRequest = Optional.ofNullable(createServiceRequest);
+        Service service = serviceServicesFactory.serviceCreator().doCreate(
+                nullableCreateServiceRequest.map(CreateServiceRequest::gatewayAccountIds).orElse(List.of()),
+                nullableCreateServiceRequest.map(CreateServiceRequest::serviceName).orElse(Map.of()));
```
alexbishop1, line 233: Presumably the entire `createServiceRequest` is not going to be null but the list of gateway account IDs or map of languages to service names might be, so perhaps this would be better:

```java
Service service = serviceServicesFactory.serviceCreator().doCreate(
        Optional.ofNullable(createServiceRequest.gatewayAccountIds()).orElse(List.of()),
        Optional.ofNullable(createServiceRequest.serviceName()).orElse(Map.of()));
```

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateIT.java
```diff
@@ -1,37 +1,117 @@
...
+                .body("current_go_live_stage", is("NOT_STARTED"))
+                .body("default_billing_address_country", is("GB"))
+                .body("agent_initiated_moto_enabled", is(false))
+                .body("takes_payments_over_phone", is(false))
+                .body("experimental_features_enabled", is(false))
+                .body("internal", is(false))
+                .body("archived", is(false))
+                .body("redirect_to_service_immediately_on_terminal_state", is(false))
+                .body("collect_billing_address", is(true));
+    }
```
alexbishop1, line 52: Maybe this method should go after all the tests that rely on it.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateIT.java
alexbishop1: I agree this is fine for an internal API.

alexbishop1 (changes requested):
Looks good! Just a few small suggestions.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateIT.java
oswaldquek: How do you mean? Like for other integration tests?

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
oswaldquek: I think this would be easier:
```
Service service = serviceServicesFactory.serviceCreator().doCreate(
        createServiceRequest == null ? List.of() : createServiceRequest.gatewayAccountIds(),
        createServiceRequest == null ? Map.of() : createServiceRequest.serviceName());
```
But am not very fussed here.

oswaldquek: ⬆️ Actually my suggestion wouldn't work 🙈

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceCreateIT.java
oswaldquek: Or do you mean moving this method to the end of the class...

alexbishop1: Yeah, at the end of the class or at least after all the tests that rely on it.

src/main/java/uk/gov/pay/adminusers/resources/ServiceResource.java
alexbishop1: Maybe this would work:

```java
Service service = serviceServicesFactory.serviceCreator().doCreate(
        Optional.ofNullable(createServiceRequest).map(CreateServiceRequest::gatewayAccountIds).orElse(List.of()),
        Optional.ofNullable(createServiceRequest).map(CreateServiceRequest::serviceName).orElse(Map.of()));
```

It would get rid of the need to have the `nullableCreateServiceRequest` variable at the expense of an extra `Optional.ofNullable(…)` call (the cost of which is probably absolutely trivial).

This could be slightly rearranged to surface the actual data and types involved a bit more:

```java
List<String> gatewayAccountIds = Optional.ofNullable(createServiceRequest).map(CreateServiceRequest::gatewayAccountIds).orElse(List.of());
Map<SupportedLanguage, String> serviceNames = Optional.ofNullable(createServiceRequest).map(CreateServiceRequest::serviceName).orElse(Map.of());

Service service = serviceServicesFactory.serviceCreator().doCreate(gatewayAccountIds, serviceNames);
```

alexbishop1 (approved):
Looks fine. One take it or leave it suggestion.

oswaldquek:
> Looks fine. One take it or leave it suggestion.

I'll take it, but will put it in the next PR. Have another PR incoming for this repo and would like to get this one merged into master.

PR #2661, opened 2024-09-18
PP-13030: Refactor method that assigns service and role to user

oswaldquek:
This commit implements these improvements:
- Use Java class representation of a POST body instead of JsonNode
- Use provided dropwizard and hibernate validation rather than custom validation

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/AssignServiceAndRoleRequest.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/AssignServiceAndRoleRequest.java
@@ -1,0 +1,24 @@
+package uk.gov.pay.adminusers.model;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import io.dropwizard.validation.ValidationMethod;
+
+import javax.validation.constraints.NotBlank;
+import javax.ws.rs.WebApplicationException;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AssignServiceAndRoleRequest(
+        @NotBlank(message = "Field [service_external_id] is required") String serviceExternalId,
+        String roleName) {
+
+    @ValidationMethod(message = "Field [role_name] must be one of 'admin', 'view-and-refund', 'view-only', 'view-and-initiate-moto', and 'view-refund-and-initiate-moto'")
+    public boolean isValidRoleName() {
+        try {
+            RoleName.fromName(roleName);
+            return true;
+        } catch (NullPointerException | WebApplicationException e) {
+            return false;
+        }
+    }
+}
--- a/src/main/java/uk/gov/pay/adminusers/service/ServiceRoleCreator.java
+++ b/src/main/java/uk/gov/pay/adminusers/service/ServiceRoleCreator.java
@@ -34,20 +34,20 @@ public class ServiceRoleCreator {
     }
 
     @Transactional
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/pay/adminusers/model/AssignServiceAndRoleRequest.java
```diff
@@ -0,0 +1,24 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import io.dropwizard.validation.ValidationMethod;
+
+import javax.validation.constraints.NotBlank;
+import javax.ws.rs.WebApplicationException;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public record AssignServiceAndRoleRequest(
+        @NotBlank(message = "Field [service_external_id] is required") String serviceExternalId,
+        String roleName) {
```
james-peacock-gds, line 13: I'm not sure why this can't just have type `RoleName` with `@Valid` annotation so you wouldn't need the `isValidRoleName` method but no doubt there are reasons.

src/main/java/uk/gov/pay/adminusers/service/ServiceRoleCreator.java
```diff
@@ -34,20 +34,20 @@ public ServiceRoleCreator(UserDao userDao, ServiceDao serviceDao, RoleDao roleDa
...
         Optional<UserEntity> userMaybe = userDao.findByExternalId(userExternalId);
-        if (!userMaybe.isPresent()) {
+        if (userMaybe.isEmpty()) {
             return Optional.empty();
         }
 
         Optional<ServiceEntity> serviceMaybe = serviceDao.findByExternalId(serviceExternalId);
-        if (!serviceMaybe.isPresent()) {
+        if (serviceMaybe.isEmpty()) {
             throw serviceDoesNotExistError(serviceExternalId);
```
james-peacock-gds, line 45: I'm not used to seeing exceptions as static references, I would always expect them to be a class, based on what we see in connector

src/test/java/uk/gov/pay/adminusers/resources/UserResourceAssignServiceAndRoleToUserIT.java
```diff
@@ -31,19 +34,16 @@ void setUp() throws Exception {
     }
     
     @Test
-    void shouldSuccess_whenAddServiceRoleForUser() {
+    void should_assign_service_and_role_to_user_successfully() {
```
james-peacock-gds, line 37: ah, yet another test naming style - this is the first time I had seen a test_name_with_underscores_between_every_word but I see it's used frequently in adminusers codebase!

james-peacock-gds (approved):
LGTM

src/main/java/uk/gov/pay/adminusers/model/AssignServiceAndRoleRequest.java
oswaldquek: I originally went with using a custom deserialiser to map the string into a `RoleName` but Jersey returns a non-422 status code. Also with the custom deserialiser we'd need a custom exception class for the response body to contain `"Field [role_name] must be one of 'admin', 'view-and-refund'...."` and it's quite nice we can define that message in this class.

src/test/java/uk/gov/pay/adminusers/resources/UserResourceAssignServiceAndRoleToUserIT.java
oswaldquek: that might be because of me 😆

PR #2781, opened 2025-04-02
ZD-6080620 Add rnngroup.ac.uk domain to email allowlist

DomBelcher:
## WHAT YOU DID
 - allow users from `rnngroup.ac.uk` to create test accounts


manishlad-cs (approved):
Additional entries look good to me.

PR #2784, opened 2025-04-08
PP-13651 refine Dropwizard 4 migration

SandorArpa:
## WHAT YOU DID

Fine tune Dropwizard 4 bom dependencies.


alexbishop1 (approved):
This changes us from using `jakarta.xml.bind-api` version 4.0.2 to version 3.0.1 but that has everything we need it and we benefit from the version being managed by Dropwizard.

PR #3310, opened 2026-07-17
PP-15596 Add migration to create service features table

ralphm10:
## WHAT YOU DID

- Add migration to create `service_features` table

### DIFF
--- a/src/main/resources/migrations/00090_create_table_service_features.sql
+++ b/src/main/resources/migrations/00090_create_table_service_features.sql
@@ -1,0 +1,10 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_service_features
+CREATE TABLE service_features (
+    id         SERIAL PRIMARY KEY,
+    service_id INT NOT NULL REFERENCES services (id) ON DELETE CASCADE,
+    feature    VARCHAR(255) NOT NULL
+);
+
+--rollback drop table service_features;


src/main/resources/migrations/00090_create_table_service_features.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_service_features
+CREATE TABLE service_features (
+    id         SERIAL PRIMARY KEY,
+    service_id INT NOT NULL REFERENCES services (id) ON DELETE CASCADE,
```
DomBelcher, line 6: does this need to explicitly declared as a foreign key?

ralphm10: I checked this myself as wasn't sure - and apparently not. Using `REFERENCES` is shorthand and the result is the same. We would only need to use the explicit one if we needed to reference multiple columns. 

See https://stackoverflow.com/questions/8595695/what-is-difference-between-foreign-key-and-reference-key

It is also used in other migrations: https://github.com/alphagov/pay-adminusers/blob/ef222539b5b7bb2c21d6629310f0c3a80ea290be/src/main/resources/migrations/00042_create_table_service_names.sql#L6

DomBelcher (approved):
LGTM

PR #3313, opened 2026-07-20
PP-15597 Add features list in service api response

ralphm10:
## WHAT YOU DID

- Add features list in service api response
- Update open api spec
- Add assertions to existing integration tests to verify that empty/populated feature lists are returned
- Some minor refactoring

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Service.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Service.java
@@ -354,7 +354,7 @@ public class Service {
         this.serviceFeatures = serviceFeatures;
     }
 
-    @ArraySchema(schema = @Schema(example = "payment links"))
+    @ArraySchema(schema = @Schema(example = "payment_links"))
     public List<String> getServiceFeatures() {
         return serviceFeatures;
     }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/adminusers/model/Service.java
```diff
@@ -347,4 +349,13 @@ public ZonedDateTime getWentLiveDate() {
     public void setCreatedDate(ZonedDateTime createdDate) {
         this.createdDate = createdDate;
     }
+
+    public void setServiceFeatures(List<String> serviceFeatures) {
+        this.serviceFeatures = serviceFeatures;
+    }
+
+    @ArraySchema(schema = @Schema(example = "payment links"))
```
DomBelcher: While there's no reason the feature names can't have spaces, I would probably expect them to be `snake_case`, but I don't really have any strong justification for it, just vibes

The best I can come up with is they can directly match names in an enum, so e.g. `PAYMENT_LINKS` could correspond to the string `payment_links` 

Even then, the examples don't really matter, they can just set the expectation of what the feature names look like

ralphm10: Updated example and test values

DomBelcher (approved):
Very nice work 🎉

PR #3322, opened 2026-07-24
PP-15598-update_patch_endpoint_to_accept_add_and_remove_feature_operation

ralphm10:
## WHAT YOU DID

- Integration tests added for both adding and removing a feature
- Added to existing parameterised validator unit tests to include validation/error handling for feature patch body 
- Amended `ServiceFeatures` to HashSet from List to ensure no duplicate features are added 
- Add `updateFeature` method to cater for both add and remove operations
- Updated `addService` to also take a Set so features can be passed to method and inserted as a separate SQL query. This led to updating all existing calls of addService to take an `emptySet()`.
- Refactored clunky if/else block in `ServiceUpdateOperationValidator` to use switch statement 
- Introduce `isValidEnumValue` to validate values on Feature enum and rename existing method to `isValidEnum` 
- Add features to `serviceDBFixture` 
- Updated openAPI spec

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Feature.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Feature.java
@@ -1,0 +1,15 @@
+package uk.gov.pay.adminusers.model;
+
+public enum Feature {
+    TEST_FEATURE("test_feature");
+
+    private final String value;
+    
+    Feature(String feature) {
+        this.value = feature;
+    }
+
+    public String getValue() {
+        return value;
+    }
+}
--- a/src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateMerchantDetailsIT.java
+++ b/src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateMerchantDetailsIT.java
@@ -10,6 +10,7 @@ import java.util.Map;
 
 import static io.restassured.http.ContentType.JSON;
 import static java.lang.String.format;
+import static java.util.Collections.emptySet;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.hasKey;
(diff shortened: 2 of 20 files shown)


src/main/java/uk/gov/pay/adminusers/model/Feature.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.adminusers.model;
+
+public enum Feature {
+    TEST_FEATURE("test_feature");
```
ralphm10, line 4: Actual features will be added here once we know what they are.

src/test/java/uk/gov/pay/adminusers/resources/ServiceResourceUpdateMerchantDetailsIT.java
```diff
@@ -25,7 +26,7 @@ public class ServiceResourceUpdateMerchantDetailsIT extends IntegrationTest {
     public void shouldSuccess_whenUpdatingMerchantDetails() throws Exception {
         String serviceExternalId = randomUuid();
         Service service = Service.from(randomInt(), serviceExternalId, new ServiceName("existing-name"));
-        databaseHelper.addService(service, randomInt().toString());
+        databaseHelper.addService(service, emptySet(), randomInt().toString());
         Map<String, Object> payload = Map.of(
```
DomBelcher, line 24: Would it be better to have the the `service` object be created with no features by default, and it be possible to set the features on this object, and then have these persisted in the database when calling `databaseHelper.addService`?

ralphm10: We weren't keen on passing an `emptySet()` to `databaseHelper.addService` either. 

What we have done is refactored the tests so that any calls to `databaseHelper.addService` are replaced by calls to `serviceDbFixture(databaseHelper)`, which creates a service with default values (and no features), or can be called with `withSomeField()` as required.  
This means setup code has been reduced in some integration tests, and now `databaseHelper.addService` is only called by `serviceDbFixture`

DomBelcher: Nice, that seems to work well

DomBelcher (approved):
Nice work!

PR #3342, opened 2026-08-11
PP-15686 Add GOVUK_PSP_IS_ADYEN feature to enable selecting Adyen in go live journey

christiandutton-mt:
## WHAT YOU DID
- Added `GOVUK_PSP_IS_ADYEN` to `Feature` enum and removed test features
- Added `getValidValues` to `Feature` enum and used in testing so that in the future new features only need adding to Feature enum
- Updated tests, schema and spec

### DIFF
--- a/src/main/java/uk/gov/pay/adminusers/model/Feature.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Feature.java
@@ -1,5 +1,8 @@
 package uk.gov.pay.adminusers.model;
 
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
 public enum Feature {
     GOVUK_PSP_IS_ADYEN("govuk_psp_is_adyen");
 
@@ -11,5 +14,11 @@ public enum Feature {
 
     public String getValue() {
         return value;
+    }
+    
+    public static String getValidValues() {
+        return Arrays.stream(Feature.values())
+                .map(Feature::getValue)
+                .collect(Collectors.joining(", "));
     }
 }
--- a/src/main/java/uk/gov/pay/adminusers/model/Feature.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Feature.java
@@ -1,8 +1,5 @@
 package uk.gov.pay.adminusers.model;
 
-import java.util.Arrays;
-import java.util.stream.Collectors;
-
 public enum Feature {
     GOVUK_PSP_IS_ADYEN("govuk_psp_is_adyen");
 
@@ -14,11 +11,5 @@ public enum Feature {
 
     public String getValue() {
         return value;
-    }
-    
-    public static String getValidValues() {
-        return Arrays.stream(Feature.values())
-                .map(Feature::getValue)
-                .collect(Collectors.joining(", "));
     }
 }
--- a/src/main/java/uk/gov/pay/adminusers/model/Feature.java
+++ b/src/main/java/uk/gov/pay/adminusers/model/Feature.java
@@ -1,8 +1,7 @@
 package uk.gov.pay.adminusers.model;
 
 public enum Feature {
-    TEST_FEATURE("test_feature"),
-    TEST_FEATURE_2("test_feature_2");
+    GOVUK_PSP_IS_ADYEN("govuk_psp_is_adyen");
 
     private final String value;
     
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/adminusers/model/Feature.java
```diff
@@ -13,4 +15,10 @@ public enum Feature {
     public String getValue() {
         return value;
     }
+    
+    public static String getValidValues() {
```
ralphm10: in `src/main/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidator.java` we have defined a constant for `FEATURES` which does something similar

src/test/java/uk/gov/pay/adminusers/resources/ServiceUpdateOperationValidatorTest.java
```diff
@@ -220,7 +221,9 @@ void replaceShouldFailWhenStringValueIsTooLong(String path, int expectedMaxLengt
     
     @Test
     void addShouldFailWhenFeatureIsInvalid(){
-        shouldFail("feature", "add", "invalidFeature", "Field [value] must be one of [test_feature, test_feature_2]");
+        var expectedErrorMessage = String.format("Field [value] must be one of [%s]", Feature.getValidValues());
```
ralphm10: if we make that constant public, we can import and use it here, something like 
`shouldFail("feature", "add", "invalidFeature", "Field [value] must be one of " + FEATURES);`

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceFindTest.java
```diff
@@ -99,9 +99,8 @@ void shouldGet_existingServiceById_withDefaultEnNameVariant() {
         assertThat(json.get("collect_billing_address"), is(serviceEntity.isCollectBillingAddress()));
         assertThat(json.get("default_billing_address_country"), is(serviceEntity.getDefaultBillingAddressCountry()));
         assertThat(json.get("current_go_live_stage"), is(String.valueOf(serviceEntity.getCurrentGoLiveStage())));
-        assertThat(json.getMap("service_features"), aMapWithSize(2));
-        assertThat(json.get("service_features.test_feature.enabled"), is(false));
-        assertThat(json.get("service_features.test_feature.enabled"), is(false));
+        assertThat(json.getMap("service_features"), aMapWithSize(1));
```
ralphm10: again if we import `FEATURES` we can do something like `assertThat(json.getMap("service_features"), aMapWithSize(FEATURES.size());` which means we don't have to update these as we add features

ralphm10 (commented):
Looks good - just a couple of suggestions

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceFindTest.java
```diff
@@ -189,9 +188,8 @@ void shouldFind_existingServiceByGatewayAccountId() {
         assertThat(json.get("sector"), is("police"));
         assertThat(json.get("internal"), is(false));
         assertThat(json.get("archived"), is(false));
-        assertThat(json.getMap("service_features"), aMapWithSize(2));
-        assertThat(json.get("service_features.test_feature.enabled"), is(true));
-        assertThat(json.get("service_features.test_feature_2.enabled"), is(false));
+        assertThat(json.getMap("service_features"), aMapWithSize(1));
```
ralphm10: as above - we could even do something like `var featureCount = FEATURES.size()` and then use that in the assertions if that is clearer

src/test/java/uk/gov/pay/adminusers/unit/service/ServiceResourceFindTest.java
```diff
@@ -42,11 +42,13 @@
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
 import static uk.gov.pay.adminusers.app.util.RandomIdGenerator.randomUuid;
+import static uk.gov.pay.adminusers.resources.ServiceUpdateOperationValidator.FEATURES;
 
 @ExtendWith(MockitoExtension.class)
 @ExtendWith(DropwizardExtensionsSupport.class)
 class ServiceResourceFindTest extends ServiceResourceBaseTest {
 
+    public static final int featureCount = FEATURES.size();
```
ralphm10: this can (should) be `private` 
Also my IDE suggested it should be a constant too, which makes sense. So named `FEATURE_COUNT`

ralphm10 (approved):

PR #3343, opened 2026-08-13
PP-15699: bump jetty

AndrzejProchyra:
* Adds Jetty BOM to override Jetty versions from `io.dropwizard:dropwizard-dependencies` to version **12.1.10**
* Instructs Dependabot to ignore the BOM for version updates


AndrzejProchyra:
This was achieved through alphagov/pay-java-commons#1455, since pay-adminusers does not need Jetty at all.

